Issue #5, opened 2017-01-25
Initial configuration of build


rmitin:
Build setup, how to build and deploy descriptions

PR #15, opened 2017-11-21
#5 Update site configuration

rmitin:
Features for update site
	- runtime feature
	- SDK feature

Packaging configuration for product job

Issue: #5
Change-Id: I8d6d6448c98317bf2f457215551607ede3d8b17a


oronzo-solazzo:
There is no scm section in the parent's pom.xml. It should not be a problem if we create release the same way we do it now.

rmitin:
@oronzo-solazzo , int this step we will only use our private Jenkins to create releases the same way as we do it now.

Issue #22, opened 2017-12-17
SCA configuration


kai-ho:
Configuration of Checkstyle, PMD, FindBugs is currently written to be compatible with Maven reporting targets and Jenkins plug-ins of that tools. This has to be adjusted to work with "plain" Maven build, i.e. violations should be logged to the console and fail the build. SCA tools will also need to be executed in the (Maven/Travis) build which requires existing SCA markers to be cleaned up.

kai-ho:
@rmitin: This issue is fixed with https://github.com/kai-ho/dsl-devkit/commit/faa30a5485c128f00d6980b8b8b50f664b8b9a2a, however the build fails on that as we have some SCA violations.  Let's fix them and I will create a pull request then...

kai-ho:
This is basically fixed with the integration of pull request https://github.com/dsldevkit/dsl-devkit/pull/25. However, PMD and Findbugs are not active yet (see https://github.com/dsldevkit/dsl-devkit/commit/3cc5e84bdfa81c8c02658272df0211d4f41356fa).

joaodinissf:
Closing — verified fixed.

PR #27, opened 2017-12-21
Issue #22 - SCA configuration and PMD cleanup

splittie96:


com.avaloq.tools.ddk.check.core.test/pom.xml
```diff
@@ -9,8 +9,4 @@
 
   <artifactId>com.avaloq.tools.ddk.check.core.test</artifactId>
   <packaging>eclipse-test-plugin</packaging>
-  <properties>
-    <pmd.ruleset>${pmd.test.ruleset}</pmd.ruleset>
-    <checkstyle.configLocation>${checkstyle.test.configLocation}</checkstyle.configLocation>
-  </properties>
```
kai-ho: Why are these properties removed? This is our pattern to distinguish between normal and test settings...

ddk-parent/pom.xml
```diff
@@ -34,6 +34,7 @@
     <compiler.showWarnings>true</compiler.showWarnings>
     <pmd.ruleset>${workspace}/ddk-configuration/pmd/ruleset.xml</pmd.ruleset>
     <pmd.test.ruleset>${workspace}/ddk-configuration/pmd/ruleset-test.xml</pmd.test.ruleset>
+    <pmd.cpd.min>100000</pmd.cpd.min>
```
kai-ho: I would add a comment why we are setting this that high and when it will be lowered again to what value (which is sensible?)

ddk-configuration/pmd/ruleset-test.xml
```diff
@@ -1,240 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
kai-ho: I would not remove this file even if the settings were the same like the normal ones. We often had cases where we wanted to have differences between normal and test code.

ddk-configuration/checkstyle/avaloq-test.xml
```diff
@@ -1,156 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
kai-ho: I would not remove this file even if the settings were the same like the normal ones. We often had cases where we wanted to have differences between normal and test code. You have an example below that should be configured in here.

com.avaloq.tools.ddk.check.core.test/pom.xml
splittie96: We discussed the options (Roman and I) and cam to the conclusion that we actually do 1 check of the entire workspace with the non-test settings, to then scan again with the test settings in each plugin. Failure happens either way. We decided to remove the test settings and just cleanup the PMD to match.

ddk-parent/pom.xml
splittie96: This is set to such a high number to "ignore" duplicate code in the scan from parent pom. When scanning the child plugins we can optionally set this to a sensible value. The exclude root value doesn't actually work when checking from parent pom so we evaluate generated code and fail...

ddk-configuration/checkstyle/suppressions.xml
```diff
@@ -5,6 +5,8 @@
     "http://www.puppycrawl.com/dtds/suppressions_1_1.dtd">
 
 <suppressions>
-    <suppress files="Messages.java"
+    <suppress files="Messages\.java"
               checks=".*"/>
+    <suppress files=".*Test.*\.java|.*Bug.*\.java"
+              checks="AvoidStarImport|EmptyBlock|IllegalCatch|JavadocMethod|JavadocType|JavadocStyle|MagicNumber|MethodName"/>
```
kai-ho: This should rather be configured in the checkstyle test ruleset (by removing modules and tuning modules configuration).  This way unwanted checks are not executed at all while in the config proposed here checks are executed but the produced markers are just suppressed (less efficient and conceptually wrong IMHO).

ddk-configuration/.fbprefs-test
```diff
@@ -0,0 +1,182 @@
+#FindBugs User Preferences
```
kai-ho: Is this really necessary? The problem with this is that Findbugs cannot use the same config for the Maven and Eclipse plug-in. So what you are doing here will only affect the Eclipse plug-in while the Maven plug-in just uses the config from the parent pom. The settings we had so far was more or less consistent. If we also want to have different FIndbugs settings for normal and test code we have to do a bit more to get this consistent.

com.avaloq.tools.ddk.check.core.test/pom.xml
kai-ho: Why's that, where can I see it? If it is caused by the aggregate option I activated for PMD I would rather remove this one.

ddk-parent/pom.xml
kai-ho: I guess we should get rid of scanning on the parent and I have a strong guess that I introduced this with PMDs aggregate option which in this case has less benefits than drawbacks. Do you like to try to remove it and check whether this leads to better results?

In general the parent should just keep the configuration for the modules and do not trigger anything by itself. Also we should avoid setting config that makes no sense for the child modules and would need to be overwritten on any child module.

splittie96: Ok, sounds good.  I guess that means I can also change back to using test preferences too.

com.avaloq.tools.ddk.check.core.test/pom.xml
rmitin: We can still remove the aggregate option, but Robbie's change also showed that there is not such a big difference between test config and productive config. So independently form activating or not the aggregate option we can ask a question if we need two different configs. Verify job suggests that tests pass the stricter rules.

ddk-configuration/.fbprefs-test
splittie96: I think this was then re-deleted in some later commit.  Maybe I should have squashed the 2, but I didn't separate the commits as well as I'd wanted to.

ddk-configuration/pmd/ruleset-test.xml
splittie96: In this case, I actually think that the 2 rules we ignore are useful even in tests - currently only "Shouldn't throw Exception" (true positives fixed) and "Assertions should have a message"(again, true positives fixed, and I think the tests are nicer for it)

com.avaloq.tools.ddk.check.core.test/pom.xml
splittie96: Turning aggregate off makes no difference here.  I wonder what else changed? As far as I can see, not much in the pom should have caused this.

ddk-configuration/.fbprefs-test
kai-ho: Nope, this file is finally in (I'm looking at changes from all commits). I think we should either go for separate settings (normal vs. tests) for all SCA tools or for none. In this pull request we remove test settings for Checkstyle and PMD but add them for Findbugs (Eclipse only). OK for you to remove the .fbprefs-test (incl. references from .project files)?

splittie96: Ok, removed in latest commit. As well as the project links

com.avaloq.tools.ddk.check.core.test/pom.xml
kai-ho: Summary: We agreed to not have different settings for productive and test code because our tests pass the stricter prod rules anyway and we prefer the simpler setup.

ddk-parent/pom.xml
kai-ho: Let's leave this for now and we consider CPD later but focus on PMD and Findbugs first

splittie96: Do you perhaps know how to disable CPD only?  I'd rather do that than change the value here, but nothing worked for me...

PR #29, opened 2017-12-22
Make CachingStateBasedContainerManager easier to extend

knutwannheden:
Refactors CachingStateBasedContainerManager by extracting a new
protected method getContainersForHandlesAndResource() from
getVisibleContainers() to make the latter easier to override.

### DIFF
--- a/com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/CachingStateBasedContainerManager.java
+++ b/com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/CachingStateBasedContainerManager.java
@@ -51,7 +51,8 @@ public class CachingStateBasedContainerManager extends StateBasedContainerManage
   @Override
   protected synchronized IContainer createContainer(final String handle, final IResourceDescriptions resourceDescriptions) {
     final IResourceDescriptions descriptionsKey = resourceDescriptions instanceof CurrentDescriptions2.ResourceSetAware
-        ? ((CurrentDescriptions2.ResourceSetAware) resourceDescriptions).getDelegate() : resourceDescriptions;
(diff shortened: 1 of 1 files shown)


com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/CachingStateBasedContainerManager.java
```diff
@@ -126,11 +127,26 @@ public void descriptionsChanged(final Event event) {
...
+   * @return list of containers, never {@code null}
+   */
+  protected List<IContainer> getContainersForHandlesAndResource(final List<String> handles, final IResourceDescription desc, final IResourceDescriptions resourceDescriptions) {
     List<IContainer> result = getVisibleContainers(handles, resourceDescriptions);
     if (!result.isEmpty()) {
       URI descURI = desc.getURI();
-      for (IContainer container : result) {
-        if (container.getResourceDescription(descURI) != null) {
+      for (int i = 0; i < result.size(); i++) {
+        if (result.get(i).getResourceDescription(descURI) != null) {
```
rmitin: You now rely on result being an array list.  Should we document it somewhere? getVisibleContainers can be overridden and implementers must be aware of this detail.

mrubanov: We have one IContainer for each handle - do i remember correct? If yes, this list is always small enough not to care about this

knutwannheden: We could create a follow-up and check if the list implements RandomAccess.

Issue #39, opened 2018-01-19
Remove ProxyFactoryEObjectDescription magic


knutwannheden:
The class ProxyFactoryEObjectDescription is used by ContainerQuery to perform some "magic" and should be considered legacy. It wraps IEObjectDescriptions returned by the index and will in getEObjectOrProxy() copy any user-data values to the corresponding EAttributes of the proxy, so that these can be queried by client code. That is weird, because the proxy isn't then used as a normal proxy anymore.

Additionally the Export DSL generator adds an additional user-data field if the name was derived from an EAttribute of the exported object. This was then of course also copied by ProxyFactoryEObjectDescription and this behavior should also be removed.

As a result the ContainerQuery no longer needs to wrap the Iterable returned by the index in another transforming Iterable. This also means that the checking for the "allow_lookup" user-data will disappear. But since that has never been an issue, I think this is acceptable.

PR #40, opened 2018-01-19
#39: Remove ProxyFactoryEObjectDescription magic

knutwannheden:
Removes ProxyFactoryEObjectDescription and everything that goes with it.

Also the Export DSL generator will now no longer add a user-data field
for any EAttribute backing the name of an exported object. This now
needs to be added manually, if that is required.


knutwannheden:
See also ASMD change https://review.sits.avaloq.net/#/c/35588/.

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/scoping/AbstractNameFunction.java
```diff
@@ -45,7 +49,8 @@ public AbstractNameFunction(final String representation) {
   /** {@inheritDoc} */
   @Override
   public QualifiedName apply(final IEObjectDescription from) {
-    return apply(from.getEObjectOrProxy());
+    LOG.warn("No explicit name function for description " + from.getEObjectURI() + " of type " + EcoreUtil.getURI(from.getEClass())); //$NON-NLS-1$ //$NON-NLS-2$
```
rmitin: It would be nice if we could make sure to never get the warning message. I suppose that's the next step with the split of naming functions. Practically if this warning occurs on a large scope the builder will probably end up doing nothing, but writing this into the log.

rmitin:
If it's fine for @abrahamm87 , it's fine for me.

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/scoping/AbstractNameFunction.java
knutwannheden: Yes, the plan was to only leave this in for a transition period until all downstream projects have adapted to these changes.

knutwannheden:
I had to rebase the PR, because I committed another conflicting change earlier today...

knutwannheden:
I should probably have resolved the conflicts by merging rather than rebasing or even resolved it using the GitHub UI...

Issue #44, opened 2018-01-26
Checkcfg template doesn't auto escape keywords


GrahamPearsonAvaloq:
Actual Result:
Register some catalog with a keyword (e.g. check) in the package name, and we see unresolved references, parsing failure, and generally quite broken checkcfg.  I imagine lots of customers have "check" as part of the package and therefore this has high impact. 

Expected Result:
where keywords are present, they should be escaped with ^

Steps to Reproduce:
Register some catalog with a keyword (e.g. check) in the package name, and we see unresolved references, parsing failure, and generally quite broken checkcfg.  I imagine lots of customers have "check" as part of the package and therefore this has high impact.

knutwannheden:
I probably don't quite understand the actual problem. But escape sequences should of course only appear in the serialized text form. In the strings in the AST model there should be no ^ symbols, otherwise there is a corresponding entry in the IValueConverterService missing. That is why I don't quite understand why you would change the IQualifiedNameConverter...

GrahamPearsonAvaloq:
The problem is, if a checkcfg refers to a check catalog whose fully-qualified name includes a checkcfg keyword (e.g. "catalog my.own.check.package.MyCatalog"), then you get an error e.g. "Couldn't resolve reference to CheckCatalog 'my.own'. A similar problem occurs if a check file tries to reference a grammar whose fully-qualified name includes a check keyword.

Is IQualifiedNameConverter the wrong place to filter out the escape characters?

joaodinissf:
Closing — verified fixed.

PR #45, opened 2018-01-26
#44: Checkcfg template doesn't auto escape keywords

GrahamPearsonAvaloq:
Add and use new class EscapingQualifiedNameConverter, which does auto
escape keywords.

Issue-Id: #44


GrahamPearsonAvaloq:
This pull request is on hold while I investigate whether this is the right approach.

Issue #46, opened 2018-01-30
Add support for contributions of export rules and checks to the language


rmitin:
Sometimes certain class of sources of one given DSL needs to get more support: it infers more types, it has specific constrains, it exports those inferred types to the index. These extensions cannot and should not be part of the core plugin for the language. To implement them separately a mechanism for extensions is needed.

We already have a way to contribute type model inference extensions with `com.avaloq.tools.ddk.xtext.modelinference.IAdditionalModelInferrer`. Now we generalize it for export rules and checks.

joaodinissf:
Closing due to inactivity — no movement in 5+ years.

PR #47, opened 2018-01-30
Initial support for export contributions

rmitin:
Generalize existing IAdditionalModelInferrer pattern to support
extending multiple services. For now we add services required for export
grouped together (fingerprint computer, resource description strategy,
qualified name provider) in IExportFeatureExtension.

We keep IAdditionalModelInferrer as is (only renamed to IModelInferrerFeatureExtension), but do not register it directly anymore in an extension point. Instead in extension point we register
ILanguageExtensions where we can register extensions for multiple services. Check extension will then also come into this service. Later we might need a similar thing for UI services when we are to extend
content assist.

All *FeatureExtensionService are language specific and can be injected by languages that support extensions. These services have default implementations (DefaultExportFeatureExtensionService,
DefaultModelInferrerFeatureExtensionService) which rely on global
ILanguageExtensionsService. ILanguageExtensions service is the one that may have different implementations for the standalone builder and for Eclipse OSGi use case.

Support for extensions in a DSL still done manually in the DSL. Support for that will be done in a follow up. In this change we only automate creation of extensions themselves (as this is the more common thing). We added a reference to the grammar for which extension is created and a name for export files.
Export DSL now generates the necessary extension classes, so extensions need nothing in
addition to .export file. 

For export files that are not defining an extension nothing changed.

Issue #46


rmitin:
Corresponding ASMD  change https://review.sits.avaloq.net/#/c/35732/

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/DefaultAdditionalExportService.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+    return traverse;
+  }
+
+  private <T> T returnFirst(final Function<IAdditionalExport, T> mapper) {
+    ensureInitilaized();
+    return exportContributions.stream().map(mapper).filter(Objects::nonNull).findFirst().orElse(null);
+  }
+
+  private void ensureInitilaized() {
```
knutwannheden: Typo. An alternative would also be to inject ILanguageContributionService into this method and thus eagerly initialize the "exportContributions" field.

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/modelinference/DefaultAdditionalInferrersService.java
```diff
@@ -0,0 +1,65 @@
...
+  @Inject
+  private Injector injector;
+
+  private List<IAdditionalModelInferrer> inferrers;
+
+  @Override
+  public void inferTargetModel(final EObject sourceModelElement, final IAcceptor<EObject> acceptor, final boolean isPrelinkingPhase) {
+    if (inferrers == null) {
+      List<IAdditionalModelInferrer> unsortedInferrers = Lists.newArrayList();
+      for (ILanguageContribution contribution : contributionService.getContributions(languageName)) {
```
knutwannheden: To me it would feel cleaner if getContributions() would already return injected instances, which then in getAdditionalModelInferrer() and getAdditionalExport() of course also would return injected instances. The ILanguageContirubtionService implementation can for instance do this the same way we do this in CoreModelSourcesService (and other places).

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/IAdditionalExport.java
```diff
@@ -0,0 +1,81 @@
...
+   * will export contained objects.
+   * </p>
+   *
+   * @param object
+   *          object
+   * @param acceptor
+   *          Acceptor to store retrieved qualified names
+   * @return true if this object's content should be processed
+   */
+  boolean doCreateEObjectDescriptions(final EObject object, final IAcceptor<IEObjectDescription> acceptor);
```
knutwannheden: The "do" prefix here seems a bit out of place...

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/IAdditionalExport.java
```diff
@@ -0,0 +1,81 @@
...
+import org.eclipse.emf.ecore.resource.Resource;
+import org.eclipse.xtext.naming.QualifiedName;
+import org.eclipse.xtext.resource.IEObjectDescription;
+import org.eclipse.xtext.util.IAcceptor;
+
+
+/**
+ * Extensions for export.
+ */
+public interface IAdditionalExport {
```
knutwannheden: I noticed that almost none of the methods in here are called by the DDK itself. Rather than only providing the plumbing I think it would make sense if the implementations of the corresponding DDK services were to already include the support for these extensions, so that the DSLs don't have to "hack" this. This was probably the idea to begin with, but I wonder if we shouldn't do this from the beginning.

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/contribution/ILanguageContribution.java
```diff
@@ -0,0 +1,44 @@
...
+package com.avaloq.tools.ddk.xtext.contribution;
+
+import com.avaloq.tools.ddk.xtext.modelinference.IAdditionalModelInferrer;
+import com.avaloq.tools.ddk.xtext.resource.IAdditionalExport;
+
+
+/**
+ * Extends the specified language.
+ */
+public interface ILanguageContribution {
```
knutwannheden: Detail: I notice we use both the terms "contribution" and "addition" in the type and method names and in the documentation we even often talk about "extension". Maybe we could use only one term? 

ILanguageExtensions? ILanguageExtensionServices? IModelInferrerContribution? IModelInferrerExtension?

In any case this interface here should use a plural form like "contributions" or "extensions".

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/modelinference/ExtendableInferredModelAssociator.java
```diff
@@ -23,7 +23,7 @@
 @Singleton
 public class ExtendableInferredModelAssociator extends InferredModelAssociator {
 
-  @Inject(optional = true)
+  @Inject
```
mrubanov: != null check below is not needed anymore then

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/DefaultAdditionalExportService.java
mrubanov: this is a singleton - should this method be synchronized?

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/IAdditionalExportService.java
```diff
@@ -0,0 +1,23 @@
...
+package com.avaloq.tools.ddk.xtext.resource;
+
+import com.google.inject.ImplementedBy;
+
+
+/**
+ * This service allows additional export rules for a given DSL to be defined outside of its plug-in.
+ */
+@ImplementedBy(DefaultAdditionalExportService.class)
+public interface IAdditionalExportService extends IAdditionalExport {
```
mrubanov: it looks strange to me that DefaultAdditionalExportService implements IAdditionalExport. I'd expect that the contributions should implement IAdditionalExport and the services would just look them up and inject members. However, it is already implemented this way with additional inferrers, so we can live with this.

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/DefaultAdditionalExportService.java
rmitin: This would look much better. Done.

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/modelinference/DefaultAdditionalInferrersService.java
rmitin: I agree that returning initialized services is better. Maybe I will just pass the injector into the method instead of the language name. The pattern we use in CoreModelSourceServices with the "dummy" source is necessary there as we don't have an injector. For CoreModelSourceServices and other alike I do hope we can get rid of the "dummy.EXTN" pattern and replace it with language to injector registry one day.

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/IAdditionalExport.java
rmitin: Right, from the interface point of view doesn't look right. At first I thought that keeping the name in the interface the same as the implementation called would help. But the implementation is going to be generated anyways.

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/IAdditionalExport.java
rmitin: Yep, just wanted to start with one language. Will do a bit more. I am extending the generator so at least the extension itself does not need to do the manual plumbing. The languages that allow extensions are not so many, so I would anyway prefer to have an "opt in" option for the languages to support the extensions. Sure, most of the code will have to go up to the DDK, but the switch will be in the language. Will try to balance between having a cleaner solution upfront and not blocking Ruben for too much who is waiting for this mechanism.

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/contribution/ILanguageContribution.java
rmitin: Right, plural form should be used here. Naming, yes... a bit too much. Will think.

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/modelinference/ExtendableInferredModelAssociator.java
rmitin: Sure, done.

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/DefaultAdditionalExportService.java
rmitin: Replaced with injection.

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/IAdditionalExportService.java
rmitin: Well. it's that the contributions and the service that aggregates them for the extension point in a DSL feature have the same API. I was not sure about this patter either. But duplicating the definition seems to be even worse.

com.avaloq.tools.ddk.xtext.export/src/com/avaloq/tools/ddk/xtext/export/generator/ExportGenerator.xtend
```diff
@@ -78,8 +81,10 @@ class ExportGenerator implements IGenerator {
   }
 
   def generateResourceDescriptionManager(ExportModel model, IFileSystemAccess fsa) {
-    val fileName = model.resourceDescriptionManager.toFileName
-    fsa.generateFile(fileName, ExportOutputConfigurationProvider.STUB_OUTPUT, resourceDescriptionManagerGenerator.generate(model, compilationContext, genModelUtil))
+  	if(!model.extension){
```
rmitin: tabs

rmitin: tab

com.avaloq.tools.ddk.xtext.export/src/com/avaloq/tools/ddk/xtext/export/Export.xtext
```diff
@@ -11,9 +11,11 @@
 grammar com.avaloq.tools.ddk.xtext.export.Export with com.avaloq.tools.ddk.xtext.expression.Expression
 
 import "http://www.eclipse.org/emf/2002/Ecore" as ecore
+import "http://www.eclipse.org/2008/Xtext" as xtext
 generate export "http://www.avaloq.com/tools/ddk/xtext/export/Export"
 
 ExportModel:
+	('export' (extension?='extension')? name=ID 'for' targetGrammar=[xtext::Grammar | QualifiedID])?
```
rmitin: tab

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/extension/ILanguageExtensionsService.java
```diff
@@ -9,26 +9,30 @@
...
    *
-   * @param languageName
-   *          the language name
-   * @return the collection of contributions, may be empty, never {@code null}
+   * @param injector
+   *          the injector of the language for which the contributions are queried
+   * @return the collection of language extensions, may be empty, never {@code null}
    */
-  Collection<ILanguageContribution> getContributions(String languageName);
+  Collection<ILanguageExtensions> getExtensions(Injector injector);
```
knutwannheden: I am also not fond of the "dummy.EXTN pattern", but the API is then at least better.

And of course we could easily add that missing registry for our own languages. Should be easy. Just in case Xtext never adds it...

com.avaloq.tools.ddk.xtext.export/src/com/avaloq/tools/ddk/xtext/export/generator/ResourceDescriptionStrategyGenerator.xtend
```diff
@@ -108,8 +108,13 @@ class ResourceDescriptionStrategyGenerator {
                   return «p.name»ExportSwitch.doSwitch(object);
                 }
               «ENDFOR»
+              «IF it.extension»
+              // Extension does not have to cover all EPackages of the language
+              return false;
```
knutwannheden: That maybe also means that it wouldn't work if an object handled by one extension is nested inside an object handled by another extension.

rmitin: This is true. Will have to revisit that. Should work for the Language example and for the Script Package case Ruben works on. So could be a separate change to revisit this. Sill part of this issue.

com.avaloq.tools.ddk.xtext.export/src/com/avaloq/tools/ddk/xtext/export/generator/ResourceDescriptionStrategyGenerator.xtend
```diff
@@ -108,8 +108,13 @@ class ResourceDescriptionStrategyGenerator {
                   return «p.name»ExportSwitch.doSwitch(object);
                 }
               «ENDFOR»
+              «IF it.extension»
+              // Extension does not have to cover all EPackages of the language
+              return false;
+              «ELSE»
               // TODO: generate code for other possible epackages (as defined by grammar)
               return true;
+              «ENDIF»
```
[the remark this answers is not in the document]
rmitin: It is. Was it the patch set few minutes ago where I missed few tabs? Forgot to activate AnyEdit in my workspace.

com.avaloq.tools.ddk.xtext.export/src/com/avaloq/tools/ddk/xtext/export/generator/FragmentProviderGenerator.xtend
```diff
@@ -65,6 +65,14 @@ class FragmentProviderGenerator {
           return super.getFragmentSegment(object);
         }
 
+        «IF it.extension»
+        @Override
+        protected CharSequence getFragmentSegmentFallback(final EObject object) {
+          // For export extension we must return null, so the logic will try other extensions
+          return null;
+        }
+        «ENDIF»
```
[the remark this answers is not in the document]
rmitin: Good point. Done.

com.avaloq.tools.ddk.xtext.export/src/com/avaloq/tools/ddk/xtext/export/generator/ResourceDescriptionStrategyGenerator.xtend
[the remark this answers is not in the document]
rmitin: Ah. Got it. Two more spaces within IF. Right.

rmitin:
Last commit corrects the not perfect indentation in generators Knut noticed. For the problem with handling of super-types in extensions I suggest I do in a follow-up together with moving the extension points into abstract classes (potentially one more adjustment of the export language, this time for the host plugin case). @mrubanov - what do you think?

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/extension/ILanguageExtensionsService.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.util.Collection;
+
+import com.google.inject.Injector;
+
+
+/**
+ * Global service that manages all extensions to all languages.
+ * <p>
+ * This service is implemented as a global singleton, one instance for all DSLs.
```
mrubanov: Well, strictly speaking this is not quite the case. The "scope" of singleton in guice is the Injector. Since we have injectors for all DSLs, there will be an instance per language

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/DefaultExportFeatureExtensionService.java
```diff
@@ -0,0 +1,86 @@
...
+import com.google.common.collect.Lists;
+import com.google.inject.Inject;
+import com.google.inject.Injector;
+import com.google.inject.Singleton;
+
+
+/**
+ * Default implementation relying on {@link ILanguageExtensions}.
+ */
+@Singleton
```
mrubanov: i'd say whenever the Singleton annotation is used, the class should be thread-safe. If I understand right, this isn't that important in this context, but pls be aware of this

rmitin: Initialisation is now thread safe as it is done by Guice.

knutwannheden:
LGTM

PR #48, opened 2018-01-31
#39: Follow-up change fixing NameFunctions#fromConstant()

knutwannheden:
The name function returned by NameFunctions#fromConstant() now also
works properly for IEObjectDescriptions.


knutwannheden:
The previous change causes a unit test failure in ASMD because of this bug in NameFunctions#fromConstant().

PR #53, opened 2018-02-08
Add support for suppressing parallel loading during build

knutwannheden:
Parallel loading has so far been applied unconditionally during builds.
For some resources this could however cause problems.

This is for instance the case with XSDResourceImpl. Because when an XSD
schema includes or imports other schemas these will be loaded as part of
the Resource#load() operation and thus end up being added to the
ResourceSet. The ParallelResourceLoader will however end up unloading
all these other resources again. This works well in most cases, because
the corresponding cross-references in the "main" resource will then be
proxified. In the case of XSD this doesn't quite work because some of
the cross-references have been annotated with "resolveProxies=false" and
as a result the client code would be exposed to proxies, which wouldn't
be there if the resource hadn't been loaded by the
ParallelResourceLoader.

As a solution there is now a new constant
ILazyLinkingResource2.PARALLEL_LOADING_SUPPORT. This is not a load
option, but rather the name (c.f. Guava's @Named) for a Boolean key
which can be bound in the language's injector. Parallel loading will be
suppressed whenever this key is bound to Boolean.FALSE.

An alternative would have been to let the client code deal with the
proxies by manually resolving them. But this would be error prone, as it
could easily be forgotten.


knutwannheden:
Reverted a small change. Since nobody had reviewed the change yet, I decided to amend the commit.

PR #58, opened 2018-02-16
Log warning instead of errors when loading binary resource fails

knutwannheden:
Small change to logging in DirectLinkingResourceStorageLoadable.


mrubanov:
Hmm, why is it a warning and not an error?

knutwannheden:
The exception is still propagated and should be handled by the client at some point. So we could even consider not logging anything. Also, the StorageAwareResource should try to load from source text if an exception is thrown. This is the case with Xtext 2.8.4 and later.

PR #60, opened 2018-02-22
Add AbstractValidationTest#resourceDiagnostic()

knutwannheden:
To be able to test for resource diagnostics this commit adds the methods
resourceDiagnostic() and linkingError() to AbstractValidationTest, which
can be used from an Xtend template just like e.g. diagnostic() and
error().

Along these lines we may in the future also want to add a convenience
syntaxError() method.

Also the code in AbstractValidationTest was cleaned up a little bit.


knutwannheden:
Because I only made a very small change for PMD I decided to amend and force push the branch.

PR #68, opened 2018-03-07
Provide DirectLinkingResourceStorageFacade#deleteStorage()

knutwannheden:
Adds a new method DirectLinkingResourceStorageFacade#deleteStorage()
which can be used to delete the binary storage for a given resource.
This method is used internally by saveResource() when the resource
contains resource-level errors (e.g. linking errors), just in case
client code calls saveResource() for a resource with errors.


knutwannheden:
Refer to internal issue CF-439.

PR #69, opened 2018-03-07
StackOverflowError protection for DelegatingScope

knutwannheden:
DelegatingScope implemented a very simple StackOverflowError protection
which only worked if a "scopeof" expression was the last in the chain
and delegated to itself.

This commit implements a more generic StackOverflowError protection
which also works in other cases. The logic is very complicated for what
it does, but that is because if implemented in the most straight-forward
manner the error logging itself would again result in a
StackOverflowError being raised!

It should be noted that the error is logged to assist the developer in
addressing the underlying problem with the scope definitions.

### DIFF
--- a/com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/scoping/DelegatingScope.java
+++ b/com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/scoping/DelegatingScope.java
@@ -10,18 +10,21 @@
  *******************************************************************************/
 package com.avaloq.tools.ddk.xtext.scoping;
 
+import java.util.Collections;
+
 import org.apache.log4j.Logger;
+import org.eclipse.emf.common.util.URI;
 import org.eclipse.emf.ecore.EClass;
 import org.eclipse.emf.ecore.EObject;
 import org.eclipse.emf.ecore.EReference;
+import org.eclipse.emf.ecore.EcorePackage;
 import org.eclipse.emf.ecore.resource.Resource;
+import org.eclipse.emf.ecore.util.EcoreUtil;
 import org.eclipse.xtext.naming.QualifiedName;
 import org.eclipse.xtext.resource.IEObjectDescription;
 import org.eclipse.xtext.scoping.IScope;
 import org.eclipse.xtext.scoping.IScopeProvider;
 
-import com.avaloq.tools.ddk.xtext.util.EObjectUtil;
-import com.google.common.base.Function;
 import com.google.common.collect.Iterables;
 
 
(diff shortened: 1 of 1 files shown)


com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/scoping/DelegatingScope.java
```diff
@@ -32,6 +35,101 @@
...
 
   private static final Logger LOGGER = Logger.getLogger(DelegatingScope.class);
+
+  /**
+   * Stores the scope ID to be logged by {@link #STACK_OVERFLOW_EOBJECT_DESCRIPTION}.
+   * <p>
+   * If multiple threads were to concurrently run into {@link StackOverflowError} it is possible that the threads would log the wrong scope IDs. But this is
+   * very unlikely to happen and the scope ID is only logged to assist the developer in addressing the programming error.
+   */
+  private static String stackOverflowScopeId;
```
rmitin: Maybe we could get back to this static field again at some point. Maybe ThreadLocal<String> instead of String? Can discuss for a followup.

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/scoping/DelegatingScope.java
```diff
@@ -110,16 +208,13 @@ public DelegatingScope(final String id, final IScopeProvider scopeProvider, fina
...
-          if (scope == DelegatingScope.this) { // NOPMD
-            LOGGER.error("Cyclic delegate detected in scope \"" + getId() + "\" while computing scope for " + EObjectUtil.getLocationString(from)); //$NON-NLS-1$ //$NON-NLS-2$
-            return IScope.NULLSCOPE;
-          }
-          return scope;
+      delegates = Iterables.transform(contexts.get(), o -> {
+        try {
+          return eReference != null ? scopeProvider.getScope(o, eReference, scopeName, originalResource)
+              : scopeProvider.getScope(o, eClass, scopeName, originalResource);
+        } catch (StackOverflowError e) {
```
rmitin: Maybe we could still detect it earlier than we run into a stack overflow. I could imagine that if we visit the same scope with the same object it is a cycle?

knutwannheden: Yes, it should be possible to implement the detection in AbstractPolymorphicScopeProvider using a HashSet (similar to LazyLinkingResource#getEObject(String)). I was just a bit worried about the performance impact. but we could test it and measure the performance impact.

I think it would actually be nice if Xtext allowed disabling the cyclic proxy resolution for performance reasons (it generates quite a bit of garbage for the GC).

PR #71, opened 2018-03-12
Provide better feedback when scoping to types with no 'name'

RGHenderson:
- Provide an error when scoping to a type with no name feature,
informing that a naming function must be used
 - Modify the message of the runtime exception when name feature is
missing, to better explain the situation

### DIFF
--- a/com.avaloq.tools.ddk.xtext.scope/src/com/avaloq/tools/ddk/xtext/scope/validation/ScopeJavaValidator.java
+++ b/com.avaloq.tools.ddk.xtext.scope/src/com/avaloq/tools/ddk/xtext/scope/validation/ScopeJavaValidator.java
@@ -18,6 +18,7 @@ import org.eclipse.core.runtime.Path;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.emf.ecore.EClass;
 import org.eclipse.emf.ecore.EObject;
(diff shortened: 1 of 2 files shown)


com.avaloq.tools.ddk.xtext.scope/src/com/avaloq/tools/ddk/xtext/scope/validation/ScopeJavaValidator.java
```diff
@@ -340,6 +342,23 @@ public void checkGuardDefaultExists(final ScopeRule context) {
...
   }
 
+  /**
+   * Check that naming functions are used when referencing a feature type with no name feature.
+   *
+   * @param scope
+   *          the scope definition to check
+   */
+  @Check
+  public void checkNamingFunctionPresentNoNameFeature(final ScopeDefinition scope) {
```
mrubanov: i think this validation doesn't cover the case when a naming function is defined in the 'naming' Btw, what was the use-case for this check?

knutwannheden: Just noticed this PR. My PR seems to be related: https://github.com/dsldevkit/dsl-devkit/pull/77.

Issue #72, opened 2018-03-13
Element cannot be primary source of multiple inferred model elements


knutwannheden:
The class InferredModelAssociator was based on a revision of JvmModelAssociator [1] which only allowed an AST element to be the primary source of a single inferred model element. Thus calling associatePrimary() for the same model element multiple times would end up only associating it as the primary of the last call. This bug was copied.

Additionally the data structure was not very suitable for the lookup of an inferred model element's primary source. Both of these problems were fixed in JvmModelAssociator [2].

[1] https://github.com/eclipse/xtext-extras/commit/0767eede515ad14474aaf008d57776222a91b3cb
[2] https://github.com/eclipse/xtext-extras/commit/7877735befe6b97a6decf068d9ce2641c42fe4e2

PR #73, opened 2018-03-13
#72: Fix problems with InferredModelAssociator

knutwannheden:
Allow a single source model element to be the primary source for
multiple inferred model elements and also improve the lookup performance
for the primary source of an inferred model element.

These changes to InferredModelAssociator also required corresponding
changes in DirectLinkingResourceStorageLoadable and
DirectLinkingResourceStorageWritable.

Additionally this commit fixes a bug in
DirectLinkingResourceStorageWritable which could cause incorrect
fragments to be computed for the inferred model element associations map
for inherited containment features as EStructuralFeature#getFeatureID()
was called rather than EClass#getFeatureID(EStructuralFeature). This in
turn could lead to ClassCastExceptions in
DirectLinkingResourceStorageLoadable.

### DIFF
--- a/com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/modelinference/InferredModelAssociator.java
+++ b/com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/modelinference/InferredModelAssociator.java
@@ -11,9 +11,12 @@
 package com.avaloq.tools.ddk.xtext.modelinference;
 
 import static com.google.common.collect.Lists.newArrayList;
-import static com.google.common.collect.Sets.newLinkedHashSet;
 
+import java.util.AbstractSet;
+import java.util.Collection;
 import java.util.Collections;
+import java.util.Iterator;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -29,9 +32,9 @@ import org.eclipse.xtext.resource.DerivedStateAwareResource;
 import org.eclipse.xtext.resource.IDerivedStateComputer;
 import org.eclipse.xtext.resource.XtextResource;
 import org.eclipse.xtext.util.IAcceptor;
+import org.eclipse.xtext.xbase.typesystem.util.Maps2;
 
-import com.google.common.collect.LinkedListMultimap;
-import com.google.common.collect.ListMultimap;
+import com.google.common.collect.Lists;
 import com.google.inject.Inject;
 import com.google.inject.Provider;
 import com.google.inject.Singleton;
@@ -42,6 +45,7 @@ import com.google.inject.Singleton;
  * Based on {@link org.eclipse.xtext.xbase.jvmmodel.JvmModelAssociator}
  */
 @Singleton
+@SuppressWarnings("restriction")
 public class InferredModelAssociator implements IInferredModelAssociations, IInferredModelAssociator, IDerivedStateComputer {
 
   private static final Logger LOGGER = Logger.getLogger(InferredModelAssociator.class);
@@ -56,15 +60,20 @@ public class InferredModelAssociator implements IInferredModelAssociations, IInf
    * An adapter that holds the mapping between source- and inferred-model elements.
    */
   public static class Adapter extends AdapterImpl {
-    private final ListMultimap<EObject, EObject> sourceToInferredModelMap = LinkedListMultimap.create();
+    private final Map<EObject, Set<EObject>> sourceToTargetMap = Maps2.newLinkedHashMapWithExpectedSize(40);
+    private final Map<EObject, Set<EObject>> targetToSourceMap = Maps2.newLinkedHashMapWithExpectedSize(40);
 
     @Override
     public boolean isAdapterForType(final Object type) {
       return Adapter.class == type;
     }
 
-    public ListMultimap<EObject, EObject> getSourceToInferredModelMap() {
-      return sourceToInferredModelMap;
+    public Map<EObject, Set<EObject>> getSourceToTargetMap() {
+      return sourceToTargetMap;
+    }
+
+    public Map<EObject, Set<EObject>> getTargetToSourceMap() {
+      return targetToSourceMap;
     }
   }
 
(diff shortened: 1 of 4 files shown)


dpetroff:
I wonder if it's worth having associatePrimary in the first place... the relationship we really care about is inferred -> (primary) source, to the point where we don't even have a "getPrimaryInferredElement" method in any of the interfaces I looked at...

knutwannheden:
The method you are looking for is `IInferredModelAssociations.getPrimarySourceModelElement(EObject)`.

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/modelinference/InferredModelAssociator.java
```diff
@@ -56,15 +60,20 @@
    * An adapter that holds the mapping between source- and inferred-model elements.
    */
   public static class Adapter extends AdapterImpl {
-    private final ListMultimap<EObject, EObject> sourceToInferredModelMap = LinkedListMultimap.create();
+    private final Map<EObject, Set<EObject>> sourceToTargetMap = Maps2.newLinkedHashMapWithExpectedSize(40);
+    private final Map<EObject, Set<EObject>> targetToSourceMap = Maps2.newLinkedHashMapWithExpectedSize(40);
```
mrubanov: hmm, i wouldn't introduce a dependency just for Maps2 (espectially restricted to a class). I think the same result you can achieve by using Multimap.newSetMultimap and providing a map of some size and a supplier that would create instances of your linked list based set

mrubanov: i am wondering whether we actually need targetToSource multimap. Do we have such cases when an inferred element has multiple source ones?

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/modelinference/InferredModelAssociator.java
```diff
@@ -97,69 +106,76 @@ protected Adapter getOrInstall(final Resource resource) {
...
-      ListMultimap<EObject, EObject> map = getSourceToInferredModelMap(inferredModelElement.eResource());
-      Set<EObject> sourceElements = newLinkedHashSet();
-      for (Map.Entry<EObject, EObject> entry : map.entries()) {
-        if (entry.getValue() == inferredModelElement) {
-          sourceElements.add(entry.getKey());
-        }
-      }
-      return sourceElements;
-    } else {
+    if (inferredModelElement == null) {
```
mrubanov: coming back to targetToSource multimap... This method isn't used in asmd at all...

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/modelinference/InferredModelAssociator.java
```diff
@@ -56,15 +60,20 @@
...
+    private final Map<EObject, Set<EObject>> targetToSourceMap = Maps2.newLinkedHashMapWithExpectedSize(40);
 
     @Override
     public boolean isAdapterForType(final Object type) {
       return Adapter.class == type;
     }
 
-    public ListMultimap<EObject, EObject> getSourceToInferredModelMap() {
-      return sourceToInferredModelMap;
+    public Map<EObject, Set<EObject>> getSourceToTargetMap() {
```
mrubanov: i liked the previous name a bit more...

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/modelinference/InferredModelAssociator.java
```diff
@@ -214,7 +230,70 @@ public void discardDerivedState(final DerivedStateAwareResource resource) {
...
+    }
+  }
+
+  /**
+   * Set implementation based on {@link LinkedList}.
+   *
+   * @param <E>
+   *          generic element type
+   */
+  public static class LinkedListBasedSet<E> extends AbstractSet<E> {
```
mrubanov: can you pls explain why this special set is necessary

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/modelinference/InferredModelAssociator.java
knutwannheden: Multimap.newSetMultimap() would not work, as we in at least one of the maps need the elements to be ordered. It would have to be something like a LinkedHashMultimap.

But I would actually prefer to stick with LinkedHashMap for the following reason: In commit https://github.com/eclipse/xtext-extras/commit/7877735befe6b97a6decf068d9ce2641c42fe4e2 the implementation of JvmModelAssociator was tweaked for performance and I believe also memory overhad. While I think it is probably possible to improve this even more, I think copying this implementation is a good starting point.

knutwannheden: While I agree that we probably don't have any such cases, I think changing the API of IInferredModelAssociations is beyond the scope

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/modelinference/InferredModelAssociator.java
knutwannheden: Correct. I have not checked other downstream projects. Also here the API was basically copied from IJvmModelAssociations.

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/modelinference/InferredModelAssociator.java
knutwannheden: I can change it back. I changed it so that the code would look more like the JvmModelAssociations class, in case we want to compare them again in the future.

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/modelinference/InferredModelAssociator.java
knutwannheden: This is needed because we want to have an ordered set (at least for the values of targetToSourceMap), where we know that the first

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/modelinference/InferredModelAssociator.java
```diff
@@ -215,6 +233,69 @@ public void discardDerivedState(final DerivedStateAwareResource resource) {
...
+      if (set == null) {
+        set = new LinkedListBasedSet<V>();
+        map.put(key, set);
+      }
+      set.add(value);
+    }
+  }
+
+  /**
+   * Set implementation based on {@link LinkedList}.
```
mrubanov: we could also leave some short comment regarding the origin/purpose of this class

Issue #76, opened 2018-03-15
Scope DSL should validate that required name function exists


knutwannheden:
A scope expression which doesn't specify explicit name functions will at runtime try to use a matching default name function. Unfortunately the Scope DSL doesn't validate that such a name function exists and may end up failing with a runtime error because of this.

PR #77, opened 2018-03-15
#76: Validate that required name function exists

knutwannheden:
Add a validation which checks that for simple scope expressions there is
a matching default name function if the experssion doesn't have any
explicit name functions.

### DIFF
--- a/com.avaloq.tools.ddk.xtext.scope/src/com/avaloq/tools/ddk/xtext/scope/validation/ScopeJavaValidator.java
+++ b/com.avaloq.tools.ddk.xtext.scope/src/com/avaloq/tools/ddk/xtext/scope/validation/ScopeJavaValidator.java
@@ -36,12 +36,15 @@ import com.avaloq.tools.ddk.xtext.expression.expression.Expression;
 import com.avaloq.tools.ddk.xtext.scope.ScopeUtil;
 import com.avaloq.tools.ddk.xtext.scope.scope.Extension;
 import com.avaloq.tools.ddk.xtext.scope.scope.GlobalScopeExpression;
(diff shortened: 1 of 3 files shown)


knutwannheden:
Any feedback on this PR?

com.avaloq.tools.ddk.xtext.scope/src/com/avaloq/tools/ddk/xtext/scope/validation/ScopeJavaValidator.java
```diff
@@ -277,6 +281,45 @@ public void checkScopeRuleOverriding(final ScopeModel context) {
...
+      return;
+    }
+    ScopeDefinition def = EObjectUtil.eContainer(expr, ScopeDefinition.class);
+    ScopeModel model = EObjectUtil.eContainer(expr, ScopeModel.class);
+    if (def != null && model != null) {
+      EClass scopeType = getType(def);
+      NamingSection namingSection = model.getNaming();
+      if (namingSection != null) {
+        for (NamingDefinition naming : namingSection.getNamings()) {
+          if (naming.getType() != null && isLeftMostSuperType(naming.getType(), scopeType)) {
```
mrubanov: mm, why only the left-most supertype?

knutwannheden: Because that matches what the scope name provider also does.

mrubanov: hmm, ok

PR #90, opened 2018-04-05
Xtext platform extensions for storing associations

rmitin:
So far platform was a set of resource descriptions plus issues. Issues
are not completely independent from resource descriptions as they have
to be associated with some resource description in the builder state.

The new concept of associations to derived objects (created by builder
participants) have a similar flavor - they are attached to existing
resource descriptions (only resources that are processed by the builder
may actually produce some derived objects), but are not part of the
resource description (the core part of Xtext builder doesn't care about
them much). So it is an extension.

Change-Id: I57455e12986ca16e4119a3e3737e4e82a52d3916


com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
```diff
@@ -716,6 +737,46 @@ protected IResourceDescription getSavedResourceDescription(final Map<URI, IResou
...
+   *
+   * @param buildData
+   *          The build data
+   * @return a map containing associations for objects derived from resources identified by their URIs
+   */
+  protected Map<URI, DerivedObjectAssociations> saveOldDerivedObjectAssociations(final BuildData buildData) {
+    if (derivedObjectAssociationsStore != null) {
+      Map<URI, DerivedObjectAssociations> cache = Maps.newHashMapWithExpectedSize(buildData.getToBeUpdated().size());
+      for (URI uri : Iterables.concat(buildData.getToBeUpdated(), buildData.getToBeDeleted())) {
+        if (cache.containsKey(uri)) {
```
mrubanov: !cache.containsKey(uri) and get rid of continue?

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/AbstractResourceDescriptionDelta.java
```diff
@@ -0,0 +1,52 @@
...
+
+  private final Map<Class<? extends Object>, Object> extensions = Maps.newHashMap();
+
+  /**
+   * Adds the extension data.
+   *
+   * @param data
+   *          the data
+   */
+  public void addExtensionData(final Object data) {
```
knutwannheden: IMHO it would make sense to have an explicit "key" parameter. That would allow using different implementation classes and the client calling getExtensionData() doesn't need t know which the implementation class is.

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/AbstractResourceDescriptionDelta.java
```diff
@@ -0,0 +1,52 @@
...
+
+import com.google.common.collect.Maps;
+
+
+/**
+ * Support for extensions on resource description delta.
+ */
+public abstract class AbstractResourceDescriptionDelta implements Delta {
+
+  private final Map<Class<? extends Object>, Object> extensions = Maps.newHashMap();
```
knutwannheden: I think I would probably allocate the HashMap on demand. An empty HashMap only uses around 128 bytes of memory, IIRC, but for 100K sources that would still add up to 12 MiB of memory.

com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/DerivedObjectAssociations.java
```diff
@@ -0,0 +1,83 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Avaloq Evolution AG and others.
```
knutwannheden: We should do something about this "2016" in the templates...

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/AbstractResourceDescriptionDelta.java
```diff
@@ -0,0 +1,52 @@
...
+   *
+   * @param <T>
+   *          the generic type
+   * @param clazz
+   *          the clazz of the extension data type
+   * @return the extension data, or {@code null} if not available
+   */
+  @SuppressWarnings("unchecked")
+  public <T> T getExtensionData(final Class<T> clazz) {
+    return (T) extensions.get(clazz);
```
mrubanov: what else are we going to use these 'extensions' map?

rmitin: At the moment that is the only usage. Main reason why this is "generic" is that resource descriptions know nothing about associations to derived objects. Currently those associations are meant for generators. I need some sort of a "pocket" to be able to attach additional "data" to pass along the delta.

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/AbstractResourceDescriptionDelta.java
rmitin: Can do that, but current implementation of the consumer will always add something for incremental builds. Only full builds will benefit. If this matters, I need to rethink consumer logic too.

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/AbstractResourceDescriptionDelta.java
rmitin: Sure. As Maxim pointed out I don't really need the map here. I want to attach a bit of payload in a way this class doesn't know the payload, but only the consumer (builder in this case). It simply allows to avoid API changes for builder participants as delta will carry the attached objects to the right place. Maybe some ideas how to avoid doing the generic extension thing?

com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/DerivedObjectAssociations.java
rmitin: Yes. can do for this single source in this change, but the rest we need to think.

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/AbstractResourceDescriptionDelta.java
knutwannheden: Why can't you just deal with this field being "null" in the instance methods of this class? Why would consumers have to change?

knutwannheden: OK. I see you already changed the code. That is what I meant.

PR #94, opened 2018-05-01
Make the ParallelResourceLoader slow loading timeout configurable 

dpetroff:
The slow loading time used in ParallelResourceLoader can now be
configured via the "resourceloader.slowloadingtime" property.

Also cleaned up some error messages to use formatted strings.


knutwannheden:
IMHO we as well make that configurable.

dpetroff:
Do we care about the name of the property much? E.g. fully quialified, camelcase, etc. or is it fine just as `resourceloader.slowloadingtime`

PR #95, opened 2018-05-14
Make grammar reference in check language optional

abrahamm87:
-make language attribute of validator extension point optional
-change check generator to handle absence of reference to grammar


com.avaloq.tools.ddk.check.core/src/com/avaloq/tools/ddk/check/generator/CheckGenerator.xtend
```diff
@@ -157,17 +157,21 @@ class CheckGenerator extends JvmModelGenerator {
...
       private static final String GRAMMAR_NAME = "«catalog.grammar.name»";
+      «ENDIF»
       private static final String CATALOG_FILE_PATH = "«catalog.checkFilePath»";
 
       /** {@inheritDoc} */
       public void doSetup() {
-        ICheckValidatorRegistry.INSTANCE.registerValidator(GRAMMAR_NAME, new «catalog.validatorClassName»());
-        ICheckCatalogRegistry.INSTANCE.registerCatalog(GRAMMAR_NAME, new ModelLocation(
+
+        ICheckValidatorRegistry.INSTANCE.registerValidator(«IF catalog.grammar !== null»GRAMMAR_NAME«ELSE»null«ENDIF», new «catalog.validatorClassName»());
```
rmitin: I would consider adding a second method with signature without the language
void registerValidator(final ICheckValidatorImpl validator);

with the intention to deprecate the one with language parameter over time.

Reasoning: we could aim for getting rid of language information in this registry as we don't really use it. We specify language when we bind catalog to the DSL. We need to see if we still keep the language reference in the catalogs in the long run. It is handy for simple single-DSL projects to just bind checks directly from the catalog without the need to use some check configuration file. So might be the syntax with the language does make sense, but then it should probably generate the same binding as we either do manually or with check configuration file.

Issue #103, opened 2018-05-23
CHECKCFG allows duplicate languages


RGHenderson:
Currently no validation is in place to disallow the same language being configured twice in the one configuration file. We should introduce a validation (error severity) and a quick-fix to merge the configs, deleting the duplicates.

PR #104, opened 2018-05-23
#103: CHECKCFG allows duplicate languages

RGHenderson:
- Add error on duplicate language configurations in CHECKCFG
 - Add quickfix to remove duplicates by merging their contents
 - Add tests for all CHECKCFG duplicate validations


com.avaloq.tools.ddk.checkcfg.core/src/com/avaloq/tools/ddk/checkcfg/validation/CheckCfgJavaValidator.java
```diff
@@ -306,10 +307,25 @@ public String apply(final ConfiguredCheck from) {
...
+   *
+   * @param checkCfg
+   *          the check configuration containing {@link ConfiguredLanguageValidator}'s
+   */
+  @Check
+  public void checkConfiguredLanguageUnique(final CheckConfiguration checkCfg) {
+    Set<String> existingLanguages = Sets.newHashSet();
+    checkCfg.getLanguageValidatorConfigurations() //
+        .stream() //
+        .filter(lang -> !existingLanguages.add(lang.getLanguage())) // Don't do the mapping first - it's easier having the EObject
```
dpetroff: Not quite so simple -- you need to check that either the parameters or the configured catalogs match before you can say they're not unique.

com.avaloq.tools.ddk.checkcfg.ui/src/com/avaloq/tools/ddk/checkcfg/ui/quickfix/CheckCfgQuickfixProvider.java
```diff
@@ -110,16 +118,45 @@ public void apply(final EObject element, final IModificationContext context) {
...
+   */
+  @Fix(IssueCodes.DUPLICATE_LANGUAGE_CONFIGURATION)
+  public void removeDuplicateLanguageConfiguration(final Issue issue, final IssueResolutionAcceptor acceptor) {
+    acceptor.accept(issue, Messages.CheckCfgQuickfixProvider_REMOVE_DUPLICATE_LANG_CONFIG_LABEL, Messages.CheckCfgQuickfixProvider_REMOVE_DUPLICATE_LANG_CONFIG_DESC, null, (element, context) -> {
+      final CheckConfiguration check = EcoreUtil2.getContainerOfType(element, CheckConfiguration.class);
+      final String languageName = ((ConfiguredLanguageValidator) element).getLanguage();
+      if (check != null && languageName != null) {
+        final List<ConfiguredLanguageValidator> allMatchingLanguages = check.getLanguageValidatorConfigurations().stream().filter(langName -> languageName.equals(langName.getLanguage())).collect(Collectors.toList());
+        if (allMatchingLanguages.size() <= 1) {
+          return; // Something went wrong. Validation guarantees 2.
```
dpetroff: Should probably be logged or throw invalid state exception.

com.avaloq.tools.ddk.checkcfg.ui/src/com/avaloq/tools/ddk/checkcfg/ui/quickfix/CheckCfgQuickfixProvider.java
```diff
@@ -110,16 +118,45 @@ public void apply(final EObject element, final IModificationContext context) {
...
+        }
+        for (ConfiguredLanguageValidator duplicate : IterableExtensions.tail(allMatchingLanguages)) {
+          // For each matching language after the first, merge contents into first and remove the language
+          IterableExtensions.head(allMatchingLanguages).getParameterConfigurations().addAll(duplicate.getParameterConfigurations());
```
dpetroff: You will want to apply each of the per-language properties from each block only to each of the catalogs configured in that block and remove them from the language level, otherwise you are changing the effective configuration. See use case mentioned above.

com.avaloq.tools.ddk.checkcfg.ui/src/com/avaloq/tools/ddk/checkcfg/ui/quickfix/CheckCfgQuickfixProvider.java
```diff
@@ -110,16 +118,45 @@ public void apply(final EObject element, final IModificationContext context) {
...
+          IterableExtensions.head(allMatchingLanguages).getCatalogConfigurations().addAll(duplicate.getCatalogConfigurations());
```
dpetroff: What if both blocks already had a configuration for some catalog?

RGHenderson: I migrated this issue from our internal Jira, in accepted state. Maybe @GrahamPearsonAvaloq (reporter) can comment on the value of this considering the valid use cases for duplicate language configs you mentioned.

RGHenderson:
I migrated this issue from our internal Jira, in accepted state. Maybe @GrahamPearsonAvaloq (reporter) can comment on the value of this considering the valid use cases for duplicate language configs you mentioned.

GrahamPearsonAvaloq:
I think we should raise an error if a ConfiguredLanguageValidator has the same language *and* the same parameterConfigurations as another ConfiguredLanguageValidator in the same CheckConfiguration.

We should also raise an error if a ConfiguredCatalog has the same catalog *and* the same parameterConfigurations as another ConfiguredCatalog in the same ConfiguredLanguageValidator. (And possibly do the same for ConfiguredCatalog blocks in legacyCatalogConfigurations.)

(We already raise an error if a ConfiguredCheck has the same check as another ConfiguredCheck in the same ConfiguredCatalog.)

In each case the quickfix could do a simple merge of the two blocks. This might result in a ConfiguredLanguageValidator containing two catalogConfigurations for the same catalog, or a ConfiguredCatalog containing two checkConfigurations for the same check, but there would be a check and a quickfix for each of those eventualities.

RGHenderson:
I wonder if the parameter configuration can actually depend on position in the source? I think it might, due to the inheritance of multiple levels of parameter configurations. If so, this becomes a little awkward to do while guaranteeing no side effect.

dpetroff:
The property configuration only depends on the nesting level, not the line number. See [CheckConfigurationPropertiesGenerator#putInheritedProperties](https://github.com/dsldevkit/dsl-devkit/blob/b09c9785fd32ebbb17d88f16e1dd4fc7043727b5/com.avaloq.tools.ddk.checkcfg.core/src/com/avaloq/tools/ddk/checkcfg/generator/CheckConfigurationPropertiesGenerator.java#L133)

Issue #139, opened 2018-08-16
"Slow loading of source" timeout is too strict


gghezzi:
The message is logged if a source takes more then 30 seconds to load, which is too strict and makes the message rather useless. A bigger time windows (1 minute would be more useful)

PR #142, opened 2018-08-17
#139: "Slow loading of source" timeout is too strict

gghezzi:
Instead of just reporting a warning when loading a resource takes longer
than a fixed amount of time we know report a warning if parsing and/or
inference takes longer than expected.

Internal Issue: DSL-1967

### DIFF
--- a/com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/resourceloader/ParallelResourceLoader.java
+++ b/com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/resourceloader/ParallelResourceLoader.java
@@ -58,7 +58,7 @@ import com.google.inject.name.Names;
 public class ParallelResourceLoader extends AbstractResourceLoader {
 
   private static final long MAX_WAIT_TIME = TimeUnit.SECONDS.toMillis(300);
-  private static final long SLOW_LOADING_TIME_DEFAULT = TimeUnit.SECONDS.toMillis(30);
+  private static final long SLOW_LOADING_TIME_DEFAULT = TimeUnit.SECONDS.toMillis(60);
   private static final String SLOW_LOADING_TIME_PROPERTY = "resourceloader.slowloadingtime"; //$NON-NLS-1$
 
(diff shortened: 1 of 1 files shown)


com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/linking/LazyLinkingResource2.java
```diff
@@ -52,6 +53,28 @@
...
 
+  private static final long SLOW_INFERENCE_TIME_DEFAULT = TimeUnit.SECONDS.toMillis(15);
+  private static final String SLOW_INFERENCE_TIME_PROPERTY = "resourceloader.slowinferencetime"; //$NON-NLS-1$
+  private static final long SLOW_PARSE_TIME_DEFAULT = TimeUnit.SECONDS.toMillis(15);
+  private static final String SLOW_PARSE_TIME_PROPERTY = "resourceloader.slowparsetime"; //$NON-NLS-1$
+  private static long slowInferenceTime;
+  private static long slowParseTime;
+
+  static {
+    String slowInferenceTimeString = System.getProperty(SLOW_INFERENCE_TIME_PROPERTY);
```
rmitin: I suppose most of the time the property is not set, then the return value will be null and we go over the NumberFormatException. I'd rather do something like

 private static final int ONE_SECOND = 1000;
private static int slowInferenceTime= SLOW_INFERENCE_TIME_DEFAULT;

String slowInferenceTimeString = System.getProperty(SLOW_INFERENCE_TIME_PROPERTY);
    if (slowInferenceTimeString != null) {
      try {
        loopTimeout = Integer.parseInt(slowInferenceTimeString ) * ONE_SECOND;
      } catch (NumberFormatException e) {
        LOGGER.warn("Could not parse slow inference time setting: " + slowInferenceTimeString );
      }
    }

because otherwise if you made a typo in the configuration, you'll just quietly revert to default. This is hard to see. Also having exceptions on the standard execution path is not nice. Also it's easier to specify time in seconds when configuring the tool.

gghezzi:
Following our internal discussions we decided for now to just increase the 1 minute the default 'slow load time ' value.

PR #143, opened 2018-08-21
Fix missing language attribute exceptions in CheckCatalogRegistryReader

abrahamm87:


rmitin:
Check catalogs indeed should not necessarily be bound to a language, although they could reference language, when they are written for one specific DSL.

PR #146, opened 2018-08-23
Fix encoding of Xpand templates

knutwannheden:
Most Xpand templates of the Valid DSL generator had the wrong encoding
and would thus fail during the MWE2 workflow using them. The encoding
has been changed to ISO-8859-1. For some reason ValidatorTests.xpt had
yet some other encoding (seems like it got garbeled during the Xtext
upgrade), but that has now also been changed to ISO-8859-1.

Since all projects currently share the same org.eclipse.core.resources
preferences (linked resource), this commit doesn't change that, even
though that would make more sense. At least for projects which have any
files with different encoding.


com.avaloq.tools.ddk.xtext.valid.generator/src/com/avaloq/tools/ddk/xtext/valid/generator/ValidatorTests.xpt
```diff
@@ -9,174 +9,174 @@
...
+�DEFINE generateTestValidator(ValidModel validModel) FOR Grammar�
+�FILE (getValidationPackage(this)+".Abstract"+this.name.toSimpleName()+"ValidTest").asPath()+".java" SRC_GEN_TEST-�
 /*
  * Generated by com.avaloq.tools.ddk.xtext.valid.generator.ValidValidatorFragment
  */
-package �getJavaValidatorName("").toJavaPackage()�;
+package �getJavaValidatorName("").toJavaPackage()�;
 
-import org.eclipse.xtext.testing.validation.ValidatorTester;
+import org.eclipse.xtext.junit4.validation.ValidatorTester;
```
knutwannheden: Also note this critical change.

PR #147, opened 2018-08-24
Add helper RecordUnresolvableXrefs

GrahamPearsonAvaloq:
Add an AutoCloseable helper to track whether an unresolvable cross-reference is encountered.

### DIFF
--- a/com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
+++ b/com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
@@ -72,7 +72,6 @@ import com.avaloq.tools.ddk.xtext.builder.tracing.ResourceValidationEvent;
 import com.avaloq.tools.ddk.xtext.extensions.AbstractResourceDescriptionsData;
 import com.avaloq.tools.ddk.xtext.extensions.IResourceDescriptionsData;
 import com.avaloq.tools.ddk.xtext.extensions.ResourceDescriptions2;
-import com.avaloq.tools.ddk.xtext.linking.ILazyLinkingResource2;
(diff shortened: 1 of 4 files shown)


com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
```diff
@@ -832,17 +833,16 @@ protected boolean recordDeltaAsNew(final Delta newDelta) {
...
-              toRetry.add(uri);
-              resourceSet.getResources().remove(resource);
-            } else {
-              final Delta intermediateDelta = manager.createDelta(oldState.getResourceDescription(uri), copiedDescription);
-              newState.register(intermediateDelta);
-              toBuild.add(uri);
+            try (final RecordUnresolvableXrefs markUnresolvableXrefs = new RecordUnresolvableXrefs(resourceSet, true)) {
+              final IResourceDescription copiedDescription = new FixedCopiedResourceDescription(description);
+              if (RecordUnresolvableXrefs.unresolvableXrefRecorded(resourceSet)) {
+                toRetry.add(uri);
```
rmitin: First we should derelease this re-try.

rmitin:
We maybe need a better mechanism to record unexpected attempts to resolve references during indexing. As a first step we should probably get rid of the old retry logic which DSLs should be relying on anymore. See #148. Then we need to think how to inform developers better about coding mistakes as they do inappropriate attempts to resolve cross-references. Maybe even report warnings if resolution is successful, but goes over index. Note: resolving via file system is OK. Some resolution over index is also OK: i.e. when ordering is enforced.

rmitin:
I tend to think that we should be getting rid of the logic related to marking failed attempts to resolve references in the indexing phase in favor of reporting warning on unauthorized index access during the indexing phase.
We should rather catch when a developer accidentally writes code that depends on index queries in the indexing phase. Sure, the fact that in indexing some references are unresolved may also be a source of information. We need to think a bit more about that.

PR #148, opened 2018-08-24
Derelease retry in indexing if reference resolution fails

rmitin:
DSLs should not rely on indexing order and if needed use other means to
resolve within indexing phase (i.e. include-like concepts over file
system directly).

So we are replacing an old mechanism that retries indexing of a source
if it attempts to resolve a reference in an indexing phase and fails
with a warning.

Now resource descriptions will be written AS IS in case of a failed
linking attempt. A warning log message will be reported, meaning that
DSL-implementers need to do something.

### DIFF
--- a/com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
+++ b/com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
@@ -794,15 +794,12 @@ public class MonitoredClusteringBuilderState extends ClusteringBuilderState
    *          The new index
    * @param monitor
    *          The progress monitor used for user feedback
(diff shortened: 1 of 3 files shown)


com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
```diff
@@ -794,15 +794,12 @@ protected boolean recordDeltaAsNew(final Delta newDelta) {
    *          The new index
    * @param monitor
    *          The progress monitor used for user feedback
-   * @return a List with the list of loaded resources {@link URI} in the first position and a list of {@link URI}s of resources that could not be loaded in the
-   *         second position.
+   * @return a List with the list of loaded resources {@link URI}
```
mrubanov: comment should be adjusted. It is no longer list of lists

rmitin: Done

PR #156, opened 2018-08-28
Disable Check and Format inferrers in pre-linking phase

gregdyke:
The check and format inferrers don't add anything we need to the xtext
index during the pre-linking phase (so don't need to be run), but
attempt to resolve references that are not yet in the index. Disabling
them removes the errors.


knutwannheden:
Seems like a rather "blunt" fix. Where exactly was the problem? To me it looked like all code resolving references was properly guarded.

gregdyke:
I plan to spend a bit of time investigating remaining problems. I'll try to drill down more specifically to what was causing the problem

PR #157, opened 2018-08-30
WIP sample ddk language

gregdyke:
missing:
- JavaValidator
- Tests
- Formatter
- Serializer
- UI functionality


gregdyke:
Not sure it's worth merging at this point. But you should be able to work from this while I finish the missing parts

PR #174, opened 2018-09-21
Fragment for grammar generation with support for semantic predicates

rmitin:
The following extension is supported:

A datatype grammar rule containing only one ID terminal rule can be annotated with @KeywordRule annotation provided a list of words so only these words can be accepted by this rule.

Example:

```
   /**
     * @KeywordRule(visible, invisible)
     * /
   VisibleKind returns VisibleKind:
     ID
   ;

```
 
The above rule will accept only 'visible' and 'invisible' identifiers. This rule we called a keyword rule because it is intended to replace usages of keywords which shall not be reserved words in the language. Reserved words are words that are not allowed to be used in identifiers.

The above example can therefore replace the following enumeration:

```
   enum VisibleKind :
       VISIBLE   = "visible"
     | INVISIBLE = "invisible"
   ;

```

The generator will insert validating semantic predicates in the rule and gated semantic predicates whenever needed in the grammar. Syntax error messages will be adjusted accordingly.

Also the fragment support more generic 

```
   /**
     * @SemanticPredicate(predicateName)
     * /
   Some Rule returns SomeEObject:
     anything
   ;
```

to insert validating and gated predicates.


rmitin:
Two things are still missing
 - content assist is not fully functional yet
 - binding in runtime module should be done by workflow

com.avaloq.tools.ddk.xtext.generator/src/com/avaloq/tools/ddk/xtext/generator/parser/common/PredicatesNaming.xtend
```diff
@@ -0,0 +1,31 @@
...
+package com.avaloq.tools.ddk.xtext.generator.parser.common
+
+import com.google.inject.Inject
+import org.eclipse.xtext.Grammar
+import org.eclipse.xtext.GrammarUtil
+import org.eclipse.xtext.generator.Naming
+
+class PredicatesNaming {
+
+  @Inject extension Naming naming
```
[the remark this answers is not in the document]
rmitin: Right. Done.

com.avaloq.tools.ddk.xtext.generator/src/com/avaloq/tools/ddk/xtext/generator/parser/antlr/AnnotationAwareXtextAntlrGeneratorFragment2.xtend
```diff
@@ -0,0 +1,218 @@
...
+    super.checkGrammar();
+    getGrammar().annotateGrammar
+  }
+
+  protected override doGenerate() {
+      super.doGenerate()
+      generateAbstractSemanticPredicate().writeTo(projectConfig.runtime.srcGen)
+  }
+
+  protected override generateProductionGrammar() {
```
andrewL-avlq: Not sure I understand why this method has been overridden when the implementation is the same, and presumably all the correct classes will be injected into the super class.

rmitin: productionGenerator is the local field here and is injected again as the type is different. AnnotationAwareAntlrGrammarGenerator cannot be made a subtype of the type used in the parent class (AntlrGrammarGenerator) as we need to do super... calls to the super.super type.

andrewL-avlq: Ah OK. Makes sense.

PR #176, opened 2018-09-28
Performance: Demand load model associations

knutwannheden:
Adds support for demand-loading the model associations from binary
storage. The new default mode is PROXIED_NODE_MODEL_AND_ASSOCIATIONS
which will both demand-load the node model and the model associations.

### DIFF
--- a/com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/persistence/ResourceLoadMode.java
+++ b/com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/persistence/ResourceLoadMode.java
@@ -35,6 +35,8 @@ public interface ResourceLoadMode {
 
   /**
    * Constituent to load from binary storage.
+   * <p>
+   * Note: It is important that the order of these literals reflect the order of the entries in the zip archives.
    */
   enum Constituent {
     /**
@@ -50,6 +52,10 @@ public interface ResourceLoadMode {
      * elements in the resource contents.
      */
     ASSOCIATIONS,
+    /**
+     * This represents the source text behind the node model.
+     */
+    SOURCE,
     /**
      * This represents the {@link org.eclipse.xtext.parser.IParseResult#getRootNode() node model}.
      */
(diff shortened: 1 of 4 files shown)


com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/persistence/ResourceLoadMode.java
```diff
@@ -50,6 +52,10 @@
      * elements in the resource contents.
      */
     ASSOCIATIONS,
+    /**
```
mrubanov: i don't quite understand why this Constituent is needed. It seems to be tightly coupled to node model one...

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/persistence/ResourceLoadMode.java
```diff
@@ -119,21 +125,42 @@ static ResourceLoadMode get(final Resource resource) {
...
 
   /**
-   * Load mode for demand-loading a resource's node model from binary storage. This is intended to be used in conjunction with {@link #PROXIED_NODE_MODEL}.
+   * Load mode for demand-loading a resource's node model from binary storage. This is intended to be used in conjunction with {@link #PROXIED_NODE_MODEL} and
+   * {@link #PROXIED_NODE_MODEL_AND_ASSOCIATIONS}.
    *
    * @return load mode for demand-loading a resource's node model, never {@code null}
    */
-  ResourceLoadMode ONLY_NODE_MODEL = of(c -> c == Constituent.RESOURCE || c == Constituent.NODE_MODEL ? Instruction.LOAD : Instruction.SKIP, "ONLY_NODE_MODEL");
+  ResourceLoadMode ONLY_NODE_MODEL = of(c -> c == Constituent.RESOURCE || c == Constituent.NODE_MODEL || c == Constituent.SOURCE ? Instruction.LOAD
```
mrubanov: didn't quite get why Constituent.RESOURCE is needed here. I thought we expect this part to loaded already

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/persistence/ResourceLoadMode.java
knutwannheden: I split NODE_MODEL into SOURCE and NODE_MODEL since they do correspond to separate entries in the ZIP archive and it makes the processing logic during the load more consistent this way. But we could of course have a special case for this.

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/persistence/ResourceLoadMode.java
knutwannheden: The RESOURCE constituent is somewhat of a dummy and is used by the PARSE load mode to indicate that the binary resource should be ignored. Vice versa `DirectLinkingResourceStorageLoadable` requires that the load mode specifies the LOAD instruction for RESOURCE.

I am of course open to other suggestions.

PR #198, opened 2018-12-20
Register inference dependencies in ExtendableInferredModelAssociator

mrubanov:


com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/modelinference/ExtendableInferredModelAssociator.java
```diff
@@ -10,34 +10,85 @@
...
 
 /**
  * Manager for associations between source and inferred model elements that allows for additional inference to be done by external providers (i.e. outside of a
  * language core plugin).
+ * Also this associator registers implicit inference dependencies.
+ * This is needed for those cases when a resource loads others to perform inference.
+ * Registering dependencies will ensure that the inference will be re-triggered once one (or many) of those 'other' resources are changed.
  */
 @Singleton
 public class ExtendableInferredModelAssociator extends InferredModelAssociator {
```
rmitin: The change is good, but I think it should to to the parent of this class - InferredModelAssociator. This class is meant to allow external contributions of inference logic. So make installedDerivedState final in the InferredModelAssociator and factor our the execution of inference into some new methods which everyone has to override. i.e. inferTargetModel.

mrubanov: done

PR #218, opened 2019-03-04
Prevent generation of empty toc.xml and contents.xml files for DSL internal checks

rubenporras:
Prevent generation of empty toc.xml and contents.xml files when the Check compiler is configured for DSL internal checks.

### DIFF
--- a/com.avaloq.tools.ddk.check.ui/src/com/avaloq/tools/ddk/check/ui/builder/CheckEclipseGeneratorConfigProvider.java
+++ b/com.avaloq.tools.ddk.check.ui/src/com/avaloq/tools/ddk/check/ui/builder/CheckEclipseGeneratorConfigProvider.java
@@ -13,7 +13,7 @@ package com.avaloq.tools.ddk.check.ui.builder;
 
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IStorage;
-import org.eclipse.emf.ecore.resource.Resource;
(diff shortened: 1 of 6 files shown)


com.avaloq.tools.ddk.check.ui/src/com/avaloq/tools/ddk/check/ui/builder/CheckEclipseGeneratorConfigProvider.java
```diff
@@ -35,11 +35,11 @@
   private CheckBuilderPreferenceAccess xbaseBuilderPreferenceAccess;
 
   @Override
-  public CheckGeneratorConfig get(final Resource resource) {
+  public CheckGeneratorConfig get(final URI uri) {
     CheckGeneratorConfig result = new CheckGeneratorConfig();
     IProject project = null;
-    if (resource != null) {
-      Pair<IStorage, IProject> pair = Iterables.getFirst(storage2UriMapper.getStorages(resource.getURI()), null);
+    if (uri != null) {
```
rmitin: I wonder why we had this null check before, as JavaDoc on the interface says that the parameter must not be null.

rubenporras: do you want me to submit a new one changing the javadoc or removing the check?

PR #221, opened 2019-03-14
Don't load the exported user data for computation of invalidations.

mrubanov:
FixedCopiedResourceDescription was adjusted to copy the object
fingerprint out of all the exported data. ResourceDescriptionDelta
doesn't use the user data items anymore.

### DIFF
--- a/com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/FingerprintResourceDescription.java
+++ b/com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/FingerprintResourceDescription.java
@@ -19,9 +19,6 @@ import java.util.Set;
 import org.eclipse.emf.common.util.URI;
 import org.eclipse.emf.ecore.EClass;
 import org.eclipse.emf.ecore.EObject;
-import org.eclipse.emf.ecore.InternalEObject;
-import org.eclipse.emf.ecore.resource.ResourceSet;
-import org.eclipse.emf.ecore.util.EcoreUtil;
 import org.eclipse.xtext.naming.QualifiedName;
 import org.eclipse.xtext.resource.IEObjectDescription;
 import org.eclipse.xtext.resource.IReferenceDescription;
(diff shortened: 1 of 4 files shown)


com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/FingerprintResourceDescription.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+    this.exportedObjects = builer.build();
+  }
+
+  private static IEObjectDescription createLightDescription(final IEObjectDescription original) {
+    InternalEObject result = (InternalEObject) EcoreUtil.create(original.getEClass());
+    result.eSetProxyURI(original.getEObjectURI());
+    String fingerprint = original.getUserData(IFingerprintComputer.OBJECT_FINGERPRINT);
+    Map<String, String> data = fingerprint != null ? Collections.singletonMap(IFingerprintComputer.OBJECT_FINGERPRINT, fingerprint) : Collections.emptyMap();
+    return EObjectDescription.create(original.getName(), result, data);
```
rmitin: You could save more memory by using a special IEObjectDescription which you could create just for this case. Then you wouldn't need even the singleton map. Also the qualified name could potentially be throws away if it is not used by the invalidation.

com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/FingerprintResourceDescription.java
```diff
@@ -0,0 +1,96 @@
...
+import com.avaloq.tools.ddk.xtext.resource.PatternAwareEObjectDescriptionLookUp;
+import com.avaloq.tools.ddk.xtext.resource.extensions.IResourceDescription2;
+import com.google.common.collect.ImmutableList;
+
+
+/**
+ * Lightweight resource description that contains only the data necessary for the invalidations.
+ * That is, the resource URI and a list of exported objects' descriptions with their fingerprints.
+ */
+public class FingerprintResourceDescription extends AbstractResourceDescription implements IResourceDescription2 {
```
rmitin: I think this needs more documentation about the assumptions. I.e. you assume EClass etc. to be part of the fingerprint. You also actually assume fingerprints to be used.

PR #231, opened 2019-04-05
Check memory when performing resource demand loads

mrubanov:
If there is not enough memory to load a needed resource -> abort
processing the current one, clear the resource set and try again later.

The new 'isMemoryLimitReached' method has been introduced in
IBuilderResourceLoadStrategy. It returns 'true' when there is not enough
memory.

The retry logic is essentially a revert of a part of
https://github.com/dsldevkit/dsl-devkit/pull/148

### DIFF
--- a/com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/Messages.java
+++ b/com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/Messages.java
@@ -15,7 +15,9 @@ import org.eclipse.osgi.util.NLS;
 
 //CHECKSTYLE:OFF
 public class Messages extends NLS {
+
   private static final String BUNDLE_NAME = "com.avaloq.tools.ddk.xtext.builder.messages"; //$NON-NLS-1$
+
   public static String MonitoredClusteringBuilderState_CANNOT_LOAD_RESOURCE;
   public static String MonitoredClusteringBuilderState_NO_MORE_RESOURCES;
   public static String MonitoredClusteringBuilderState_PHASE_ONE_DONE;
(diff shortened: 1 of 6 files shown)


com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
```diff
@@ -928,7 +942,10 @@ protected boolean recordDeltaAsNew(final Delta newDelta) {
         if (!loadingStrategy.mayProcessAnotherResource(resourceSet, resourceSet.getResources().size())) {
```
rmitin: || resourceToRetry == null 

I would also check if we're retrying. Asking runtime for the information is not so exact. So if we failed because of memory, we need to clear it.

com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/Messages.java
```diff
@@ -15,14 +15,18 @@
...
   private static final String BUNDLE_NAME = "com.avaloq.tools.ddk.xtext.builder.messages"; //$NON-NLS-1$
+
   public static String MonitoredClusteringBuilderState_CANNOT_LOAD_RESOURCE;
   public static String MonitoredClusteringBuilderState_NO_MORE_RESOURCES;
   public static String MonitoredClusteringBuilderState_PHASE_ONE_DONE;
   public static String MonitoredClusteringBuilderState_UPDATE_DESCRIPTIONS;
   public static String MonitoredClusteringBuilderState_WRITE_DESCRIPTIONS;
   public static String MonitoredClusteringBuilderState_WRITE_ONE_DESCRIPTION;
   public static String MonitoredClusteringBuilderState_COULD_NOT_PROCESS_DUE_TO_STACK_OVERFLOW_ERROR;
+  public static String MonitoredClusteringBuilderState_TRY_AGAIN;
```
rmitin: Not used in this change. However, logging INFO would not be bad.

com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
```diff
@@ -925,10 +940,13 @@ protected boolean recordDeltaAsNew(final Delta newDelta) {
           monitor.worked(1);
         }
 
-        if (!loadingStrategy.mayProcessAnotherResource(resourceSet, resourceSet.getResources().size())) {
+        if (!loadingStrategy.mayProcessAnotherResource(resourceSet, resourceSet.getResources().size()) || resourceToRetry == null) {
```
rmitin: || resourceToRetry != null

PR #234, opened 2019-05-08
Small improvements on the builder

rubenporras:
- In doUpdate: Calculate the memory used when linking resources only if
the tracing
event ResourceLinkingMemoryEvent has been enabled.
- In writeResources: The exception block has
been protected to not fail if URI is null. Throwable is no longer
catched (we assume that the bug in Xtext has been fixed, as it did not
appear in the last years).
- In writeNewResourceDescriptions: correctly guard exceptions to not
leave the setIndexing flag wrongly set.
- In queueAffectedResources: do not catch a exception which seems fatal
just to log a warning and continue.

### DIFF
--- a/com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
+++ b/com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
@@ -362,28 +362,27 @@ public class MonitoredClusteringBuilderState extends ClusteringBuilderState
     Collection<IResourceDescription.Delta> deltas = doClean(toBeRemovedCopy, subMonitor.newChild(1));
 
     final ResourceDescriptionsData newData = getCopiedResourceDescriptionsData();
-    ResourceDescriptionChangeEvent event = null;
+
+    checkForCancellation(monitor);
     try {
-      checkForCancellation(monitor);
       for (IResourceDescription.Delta delta : deltas) {
         newData.removeDescription(delta.getOld().getURI());
       }
-      event = new ResourceDescriptionChangeEvent(deltas);
+      ResourceDescriptionChangeEvent event = new ResourceDescriptionChangeEvent(deltas);
       checkForCancellation(monitor);
       updateDeltas(event.getDeltas(), null, subMonitor.newChild(1));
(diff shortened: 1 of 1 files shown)


com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
```diff
@@ -362,28 +362,27 @@ protected void propagateDependencyChains(final Set<URI> toBeUpdated, final IReso
...
       }
-      event = new ResourceDescriptionChangeEvent(deltas);
+      ResourceDescriptionChangeEvent event = new ResourceDescriptionChangeEvent(deltas);
       checkForCancellation(monitor);
       updateDeltas(event.getDeltas(), null, subMonitor.newChild(1));
       // update the reference
       setResourceDescriptionsData(newData, monitor);
       // CHECKSTYLE:CHECK-OFF IllegalCatch
+      notifyListeners(event);
+      return event.getDeltas();
```
mrubanov: not related to your change, but can you pls fix the typo on line 380? (CHEKC)

com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
```diff
@@ -887,16 +897,6 @@ protected boolean recordDeltaAsNew(final Delta newDelta) {
               newState.register(new DefaultResourceDescriptionDelta(oldDescription, null));
             }
           }
-          // CHECKSTYLE:CHECK-OFF IllegalCatch
-          // If we couldn't load it, there's no use trying again: do not add it to the queue
-        } catch (final Throwable e) {
```
mrubanov: i think this catch block has to be brought back... Otherwise, if some exception occurs the whole build will fail

com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
rubenporras: done

PR #235, opened 2019-05-10
Remove scope pruning

junfuller:
Change-Id: Ib2420dd1ba596be3eddd2b7b274bb08af2859127


com.avaloq.tools.ddk.xtext.scope/src/com/avaloq/tools/ddk/xtext/scope/generator/ScopeProviderGenerator.xtend
```diff
@@ -227,8 +227,6 @@ class ScopeProviderGenerator {
 
   def scopes(ScopeExpression it, ScopeModel model, String typeOrRef, ScopeDefinition scope, Boolean isGlobal) {
     val b = new StringBuilder
```
mrubanov: i guess the StringBuilder isn't really needed here now..

PR #243, opened 2019-07-17
Avoid triggering inference when executing container queries

gregdyke:
Calls to resource.getContents.get(0) accidentally trigger inference if
the resource was loaded without inference. Even in cases where the
context.eResource() and originalResource are different, we do not need
to get the an EObject from originalResource to find its visible
containers.

- Deprecate the first parameter in
ContainerQuery#execute(EObject/Resource context, Resource
originalResource) as it is never used
- Add AbstractPolymorphicScopeProvider#getVisibleContainers(XtextResource)
as alternative to
AbstractPolymorphicScopeProvider#getVisibleContainers(EObject, Resource)
as the EObject is only relevant if the resource is null.
- Refactor all methods to avoid useless calls to getContents()

### DIFF
--- a/com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/scoping/AbstractPolymorphicScopeProvider.java
+++ b/com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/scoping/AbstractPolymorphicScopeProvider.java
@@ -286,7 +286,9 @@ public abstract class AbstractPolymorphicScopeProvider extends AbstractScopeProv
   protected abstract IScope doGetScope(final EObject context, final EClass type, final String scopeName, final Resource originalResource);
 
   /**
-   * Return the visible containers given a context object.
+   * Return the visible containers given a context object and an originalResource.
+   * <p>
+   * Returns the containers visible from the originalResource, unless it is null, in which case, returns the containers visible from the context's resource
    *
    * @param context
    *          The context object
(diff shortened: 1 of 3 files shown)


com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/scoping/AbstractPolymorphicScopeProvider.java
```diff
@@ -296,32 +298,38 @@ protected boolean doGlobalCache(final EObject context, final EClass type, final
...
-    final ResourceCache<URI, List<IContainer>> cache = CacheManager.getInstance().getOrCreateResourceCache("AbstractPolymorphicScopeProvider#visibleContainers", rsc); //$NON-NLS-1$
+    final ResourceCache<URI, List<IContainer>> cache = CacheManager.getInstance().getOrCreateResourceCache("AbstractPolymorphicScopeProvider#visibleContainers", resource); //$NON-NLS-1$
     if (cache != null) {
       result = cache.get(uri);
       if (result != null) {
         return result;
       }
     }
 
-    final EObject ctx = ctxRsc != context.eResource() ? ctxRsc.getContents().get(0) : context;
```
rubenporras: The rest of the change is good for me, but I do not understand why we needed this condition or why we can remove it now.

gregdyke: I don't understand why we needed the condition either.

The only use that is made of the ctx variable is to get ctx.eResource() in lines 323 and 324. In all cases, the result of ctx.eResource() is ctxRsc (which is now named resource):

- if context.eResource() and originalResource are the same, it's the same
- if originalResource is null, it's replaced by context.eResource(), so they are the same
- if context.eResource() and originalResource are different:
  - ctxRsc == originalResource
  - ctx == originalResource.getContents().get(0)
  - ctx.eResource() == ctxRsc

PR #249, opened 2019-07-29
Update .travis.yml

gregdyke:


.travis.yml
```diff
@@ -1,5 +1,6 @@
 language: java
 sudo: false
+dist: trusty
```
rubenporras: Why do we go for this instead of changing it to openjdk8?

gregdyke: Just changing openjdk was not enough as other changes were needed to work with Xenial. I did two of those changes and then got stuck, so figured reverting to trusty would serve just as well for now

PR #260, opened 2019-09-17
Warn if database flush takes longer than thirty seconds

rubenporras:
Output a warning to the log if database flush takes longer than thirty
seconds. This will help diagnosing cases where the database write
operations are blocking the build process.

### DIFF
--- a/com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
+++ b/com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
@@ -112,6 +112,7 @@ public class MonitoredClusteringBuilderState extends ClusteringBuilderState
   public static final long CANCELLATION_POLLING_DELAY = 200; // ms
 
   public static final int STACK_TRACE_LIMIT = 10;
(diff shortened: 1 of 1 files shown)


com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
```diff
@@ -112,6 +112,7 @@
   public static final long CANCELLATION_POLLING_DELAY = 200; // ms
 
   public static final int STACK_TRACE_LIMIT = 10;
+  private static final int COMMIT_WARN_WAIT_SEC = 30;
```
mrubanov: how did you come to this figure of 30sec? I am just wondering - could we perhaps always output the time it took to commit changes? We will then also get some idea of what the usual numbers are

rubenporras: My tests have shown that the WriteBehindBuffer is typically always able to flush the buffer in the background thread, thus, by the time the builder calls flush on its own, there is nothing to do. So I think for our cases the typical number is close to zero. I have set 30 sec because I think that if the WriteBehindBuffer  is lagging behind but this delays the build less than 30 sec, it is not worth to log. Does this explain the number?

PR #269, opened 2019-10-09
Allow parallel setup of languages in the standalone setup

rubenporras:
Allow parallel setup of languages in the standalone setup by locking on
a provided object during the EMF registration.


com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/build/AbstractDynamicSetupService.java
```diff
@@ -12,34 +12,13 @@
...
-    throw new UnsupportedOperationException();
-  }
-
-  @Override
-  public Injector doSetup(final Module overrideModule, final Module... additionalModules) {
-    // this implementation can be removed once all subclasses have been regenerated using DDK 4.0.0
-    // until then it is required for backwards compatibility
-    return doSetup(overrideModule);
-  }
+  public static final Integer SETUP_LOCK = new Integer(1); // NOPMD Avoid instantiating Integer objects
```
mrubanov: kind of ok, but normally one just creates an instance of Object type for a lock

com.avaloq.tools.ddk.xtext.generator/src/com/avaloq/tools/ddk/xtext/generator/builder/StandaloneBuilderIntegrationFragment2.xtend
```diff
@@ -124,7 +132,11 @@ class StandaloneBuilderIntegrationFragment2 extends AbstractXtextGeneratorFragme
         public Injector createInjectorAndDoEMFRegistration() {
           registerEPackages();
           Injector injector = createInjector();
-          register(injector);
+          if (lock != null) {
```
mrubanov: this is a bit strange... If the lock isn't set for some reason then we don't register the injector. Could we rather make sure the lock isn't null?

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/build/AbstractDynamicSetupService.java
rubenporras: Good point. Done.

com.avaloq.tools.ddk.xtext.generator/src/com/avaloq/tools/ddk/xtext/generator/builder/StandaloneBuilderIntegrationFragment2.xtend
rubenporras: Yes, I forgot the else branch. I have it now.
There are two different execution paths, the first one goes over the AbstractDynamicSetupService, there there is always a lock object, which is created statically in AbstractDynamicSetupService and passed to the constructor.
There is a second path, typically used in core test plugins, which calls the constructor without a lock. I wanted to keep the API for this case constant and therefore the if/else. I could also create a dummy object in the constructor which takes no lock object and synchronize into it, but that might be more strange, depending on the point of view. At the end I settled for the if/else block.
Would you prefer the other one? Or would you have other suggestions?

mrubanov: Looks a bit strange to me, i'd probably unify these 2 paths... however i don't see any harm in what you did, so I leave it to you to decide

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/build/AbstractDynamicSetupService.java
```diff
@@ -12,34 +12,13 @@
...
-    throw new UnsupportedOperationException();
-  }
-
-  @Override
-  public Injector doSetup(final Module overrideModule, final Module... additionalModules) {
-    // this implementation can be removed once all subclasses have been regenerated using DDK 4.0.0
-    // until then it is required for backwards compatibility
-    return doSetup(overrideModule);
-  }
+  public static final Integer SETUP_LOCK = new Object;
```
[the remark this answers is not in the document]
rubenporras: Done.

com.avaloq.tools.ddk.xtext.generator/src/com/avaloq/tools/ddk/xtext/generator/builder/StandaloneBuilderIntegrationFragment2.xtend
rubenporras: Okay, then I would say we can leave it as it is. Thanks.

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/build/AbstractDynamicSetupService.java
```diff
@@ -12,34 +12,13 @@
...
-    throw new UnsupportedOperationException();
-  }
-
-  @Override
-  public Injector doSetup(final Module overrideModule, final Module... additionalModules) {
-    // this implementation can be removed once all subclasses have been regenerated using DDK 4.0.0
-    // until then it is required for backwards compatibility
-    return doSetup(overrideModule);
-  }
+  public static final Integer SETUP_LOCK = new Object();
```
mrubanov: Object rather instead of integer

rubenporras: Done

PR #275, opened 2019-11-08
Propagate the Source Level URIs to the child ResourceSet

rubenporras:
Propagate the Source Level URIs to the child ResourceSet, this allows
code which runs inside the ParallelResourceLoader (mainly inference) to
use the binary models.


rubenporras:
The build runs through without any exception or warning and there is no variance in the markers, so I will submit it.

PR #283, opened 2019-11-22
Improve name registration

rubenporras:


com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/linking/LinkingService.java
```diff
@@ -64,70 +65,114 @@
...
+   *
+   * @param context
+   *          context object within which a reference is being set to unresolved, must not be {@code null}
+   * @param name
+   *          the lookup name, may be {@code null}
+   * @param type
+   *          the lookup type, may be {@code null}
+   */
+  public void registerUnresolvedReference(final EObject context, final String name, final EClass type) {
+    ImportedNamesTypesAdapter adapter = getImportedNamesAdapter(context);
```
mrubanov: i guess we can factor this code out in a separate method... (everything except the 2nd line)

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/linking/LinkingService.java
```diff
@@ -176,7 +221,18 @@ protected EObject getSingleElement(final EObject context, final EReference refer
     } else {
       desc = scope.getSingleElement(qualifiedLinkName);
     }
-    return desc == null ? null : desc.getEObjectOrProxy();
+
+    EObject target = desc == null ? null : desc.getEObjectOrProxy();
+
+    if (target != null && isImportRequired(context, reference, target)) {
+      if (target.eIsProxy()) {
```
mrubanov: do you know in which cases it is proxy and in which not?

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/linking/LinkingService.java
rubenporras: Done.

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/linking/LinkingService.java
rubenporras: My understanding is that whenever we use an index lookup to find our object, we have a proxy, since we get it from the index the name of the proxy is the one which has the calculated name using the logic in the export file.
For the cases where we look within a resource (e.g. some chained feature calls), then we have resolved objects.

PR #288, opened 2019-11-28
Recalculate the qualified names if possible

rubenporras:
If possible, do not look for the qualified names in the resource
descriptor, recalculating the values is much faster, as the lookup
involve very expensive operations:
- Create a collection with all the resource descriptions of the
referenced resource (only first lookup).
- Iterate through the list every time and compare the fragments.

### DIFF
--- a/com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/linking/LinkingService.java
+++ b/com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/linking/LinkingService.java
@@ -17,9 +17,11 @@ import java.util.List;
 import org.eclipse.emf.ecore.EClass;
 import org.eclipse.emf.ecore.EObject;
 import org.eclipse.emf.ecore.EReference;
+import org.eclipse.emf.ecore.resource.Resource;
 import org.eclipse.xtext.linking.impl.DefaultLinkingService;
 import org.eclipse.xtext.linking.impl.ImportedNamesAdapter;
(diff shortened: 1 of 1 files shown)


com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/linking/LinkingService.java
```diff
@@ -82,13 +84,22 @@ private void registerNamedType(final EObject context, final QualifiedName name,
...
+      IQualifiedNameProvider nameProvider = ((LazyLinkingResource2) eResource).getService(IQualifiedNameProvider.class);
+      if (nameProvider != null) {
+        targetName = nameProvider.getFullyQualifiedName(target);
       }
+    } else {
+      final IResourceDescriptions resourceDescriptions = provider.getResourceDescriptions(context.eResource());
+      Iterator<IEObjectDescription> exports = resourceDescriptions.getExportedObjectsByObject(target).iterator();
+      if (exports.hasNext()) {
+        targetName = exports.next().getName();
+      }
```
rubenporras: I think this code will never be executed, and it is indeed never executed in a FULL build with ASMD, I did not removed it because I am not sure if the same holds for ABMD/AMI-WS, even though I think it should. What do you think, should I remove it?

mrubanov: In general we may have resources that are not LazyLinkingResource2 ones, so let's keep it

PR #299, opened 2020-01-08
Add a generic Resource Set Cache Adapter which uses Strong References

rubenporras:
Introduce a generic resource set cache adapter which uses strong
references


com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/StrongReferencerCacheAdapter.java
```diff
@@ -0,0 +1,57 @@
...
+import com.avaloq.tools.ddk.caching.MapCache;
+
+
+/**
+ * A generic resource set cache adapter which uses strong references.
+ *
+ * @param <V>
+ *          the type of the stored values
+ */
+public class StrongReferencerCacheAdapter<V> extends AdapterImpl implements IResourceSetCache<V> {
```
mrubanov: as far as i can tell the CacheAdapter class is pretty much the same - would it perhaps make sense to add some constructor parameter which would then switch to strong references? (we can use soft ones by default)

rubenporras: I will take a look tomorrow at this.

PR #300, opened 2020-01-09
Add a generic Resource Set Cache Adapter which uses Strong References

rubenporras:
Introduce a generic resource set cache adapter which uses strong
references.

In addition a method putIfAbsent and computeIfAbsent have been added to
the CacheAdapter, to offer this new Java API of the Map class also to
the users of the CacheAdapter.

### DIFF
--- a/com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/CacheAdapter.java
+++ b/com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/CacheAdapter.java
@@ -10,6 +10,8 @@
  *******************************************************************************/
 package com.avaloq.tools.ddk.xtext.resource;
 
+import java.util.function.Function;
+
 import org.eclipse.emf.common.notify.Notifier;
 import org.eclipse.emf.common.notify.impl.AdapterImpl;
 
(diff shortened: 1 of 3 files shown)


com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/StrongReferencerCacheAdapter.java
```diff
@@ -0,0 +1,32 @@
...
+import com.avaloq.tools.ddk.caching.CacheConfiguration;
+
+
+/**
+ * A generic resource set cache adapter which uses strong references.
+ *
+ * @param <V>
+ *          the type of the stored values
+ */
+public class StrongReferencerCacheAdapter<V> extends CacheAdapter<V> {
```
mrubanov: are you sure we actually need this class? Could we perhaps just pass a boolean argument to CacheAdapter to use strong references instead of soft ones? Then the consumer can decide which references he wants without inheriting from a different class

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/StrongReferencerCacheAdapter.java
```diff
@@ -0,0 +1,32 @@
...
+  @Override
+  protected CacheConfiguration getCacheConfiguration() {
```
mrubanov: be aware that in general invoking the protected/public methods from constructors/initializers isn't a safe pattern, as these methods can be overridden and invoke while the object isn't completely initialized

PR #302, opened 2020-01-09
Introduce interface IDynamicDefaultPackageSetupService

### LINKED ISSUE #301: Introduce mechanism for language independent XMI setup
It is sometimes required to initialize an XMI independent from languages.
To allow this a mechanism is required that allows a builder to discover all the required classes so it can iterate over them

gerdleon:
Through introduction of this interface a builder can get all implementations of
this service and call the "doPackageSetup" method which performs the
package initialization.

Closes #301


com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/build/IDynamicDefaultPackageSetupService.java
```diff
@@ -0,0 +1,31 @@
...
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Avaloq Evolution AG - initial API and implementation
+ *******************************************************************************/
+
+package com.avaloq.tools.ddk.xtext.build;
+
+/**
+ * This interface is used when a default package needs to be initialized.
```
mrubanov: what is meant by default package in this context? What is a non-default then?

PR #303, opened 2020-01-14
Optimize DefaultCheckImpl.internalValidate

mrubanov:
- The 'methodsForType' field is now a usual map instead of a cache with
weak references. The cached MethodWrapper object do not seem to be
extremely big so it makes sense to compute them once and re-use, as the
internalValidate method gets invoked really often.

- Iteration through the list of the methods is now done using simple for
loop, without creation of an iterator

- Computation of 'methodsForType' moved toward the beginning of the
method, so we don't need to create 'State' if there it isn't needed.

### DIFF
--- a/com.avaloq.tools.ddk.check.runtime.core/src/com/avaloq/tools/ddk/check/runtime/issue/DefaultCheckImpl.java
+++ b/com.avaloq.tools.ddk.check.runtime.core/src/com/avaloq/tools/ddk/check/runtime/issue/DefaultCheckImpl.java
@@ -15,7 +15,6 @@ import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


com.avaloq.tools.ddk.check.runtime.core/src/com/avaloq/tools/ddk/check/runtime/issue/DefaultCheckImpl.java
```diff
@@ -112,10 +112,24 @@ private void initCheckMethodCache() {
    *          the context
    * @return true, if successful
    */
-  protected final boolean internalValidate(final EClass class1, final EObject object, final DiagnosticChain diagnostics, final Map<Object, Object> context) {
+  protected final boolean internalValidate(final EClass class1, final EObject object, final DiagnosticChain diagnostics, final Map<Object, Object> context) { // NOPMD
```
rubenporras: what is the warning we are suppressing with NOPMD?

mrubanov: NPathComplexity

PR #304, opened 2020-01-20
Fix for NullPointerException in DirectLinkingResourceStorageLoadable. 

markdomeng:
initialization done by calling resource.getContents() and resource.eAdapters()


markdomeng:
> I think we should remove the condition
> if (e instanceof NullPointerException) { // NOPMD

done

rubenporras:
Hi Mark,

could you improve the commit message so that someone in the future would understand the problem? I think we should also include a link to the bug we reported to Xtext in the commit message.

PR #307, opened 2020-01-24
Only store binary resources if platform has a binary model store.

andrewL-avlq:


com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
```diff
@@ -208,6 +208,11 @@
    */
   private final IXtextTargetPlatformManager targetPlatformManager;
 
+  /**
+   * During startup (and maybe in other cases) it is possible for us to be given a default platform which has no binary storage.
+   */
+  private boolean isBinaryModelStorageAvailable;
```
mrubanov: probably not a very good idea. What i think happens is that when one restarts ice (and there was an editor opened before restarting) activation of the opened editor triggers a default xtext platform (without binary store) to be created and therefore then we see the NPEs. If you now cache the information about the absence of the binary model store it won't be ever changed since MonitoredClusteringBuilderState is a singleton. This will essentially switch off the binary models.

What i think we should investigate is why don't we re-create an xtext platform (with binary models support) before we start the workspace build.

PR #320, opened 2020-04-16
Support Case Sensitive and Insensitive fragments

rubenporras:
Support Case Sensitive and Insensitive fragments by adding a new marker
interface EClass ICaseSensitiveNamedElemnt.

By default, when selecting the fragments, it is considered that the
fragment is case insensitive unless the EClass inherits from the new
marker interface.

Change-Id: I81cd2654089d0a320a5632d2f7054b6cd5a84c46

### DIFF
--- a/com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/AbstractSelectorFragmentProvider.java
+++ b/com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/AbstractSelectorFragmentProvider.java
@@ -10,12 +10,16 @@
  *******************************************************************************/
 package com.avaloq.tools.ddk.xtext.resource;
 
+import java.util.concurrent.ConcurrentMap;
+
 import org.eclipse.emf.common.util.EList;
+import org.eclipse.emf.ecore.EClass;
 import org.eclipse.emf.ecore.EObject;
 import org.eclipse.emf.ecore.EStructuralFeature;
 
 import com.avaloq.tools.ddk.xtext.linking.AbstractFragmentProvider;
 import com.avaloq.tools.ddk.xtext.linking.ShortFragmentProvider;
+import com.google.common.collect.Maps;
 import com.google.inject.Inject;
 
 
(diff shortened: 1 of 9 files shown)


com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/AbstractSelectorFragmentProvider.java
```diff
@@ -118,6 +124,10 @@ public boolean appendFragmentSegment(final EObject object, final StringBuilder b
     return appendFragmentSegmentFallback(object, builder);
   }
 
+  private Boolean isCaseSensitive(final EClass eClass) {
+    return (Boolean) eclassToCaseSensitive.computeIfAbsent(eClass, k -> eClass.getEAllSuperTypes().stream().map(EClass::getName).anyMatch(n -> "ICaseSensitiveNamedElement".equals(n))); //$NON-NLS-1$
```
mrubanov, line 128: I'd rather do something like TypemodelPackage.Literals.ICASE_SENSITIVE_NAMED_ELEMENT.isSupertypeOf(eClass), which should be nicer and probably faster

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/AbstractSelectorFragmentProvider.java
```diff
@@ -38,6 +42,8 @@
   @Inject
   private ShortFragmentProvider shortFragmentProvider;
 
+  private final ConcurrentMap<Object, Object> eclassToCaseSensitive = Maps.newConcurrentMap();
```
mrubanov, line 45: don't know whether caching will help a lot here... i guess a usual hashmap will do here if you want to keep it

rubenporras: Done

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/AbstractSelectorFragmentProvider.java
rubenporras: This cannot be done, then a cycle in the MANIFEST dependencies appears

PR #352, opened 2020-11-23
Regenerate after xtext2 18 upgrade

rubenporras:


gregdyke (approved):
Went through everything to sanity check there was nothing unexpected

PR #366, opened 2021-02-15
Fix to proxy URI fragments when unloading the model from a resource.

andrewL-avlq:
Unloading the model sets the proxy URI fragments on the model elements.
Added override to LazyLinkingResource2 that sets the fragments before
the model elements are removed from the resource so that the proxy URIs
are created using the correct URI fragment provider.


rubenporras:
Is this a bug in Xtext or we just need the fix because some speciallity of the DDK?
Could we comment the code to make more clear what lines changed? It will help in the future to know if the override is still neede.

andrewL-avlq:
I would argue that it is a bug in Xtext - it's implementation uses a fallback uri fragment provider which appears to know nothing about the custom providers we have added to the resource.

rubenporras (approved):

PR #372, opened 2021-03-11
added a guard to prevent calling reference.isMany() when reference is null

markdomeng:
NPE occurs when the ShortFragmentProvider hits a StringLiteral which has no reference object (this happens when the source is setup like the reproducer).


markdomeng (commented):
hi @rubenporras , could you take a look at this change? (for some reason I cant assign a reviewer)

enzosanmiguel (approved):

PR #387, opened 2021-04-28
Do not index derived or transient cross-references

rubenporras:
Do as the Xtext base implementation and do not index derived or
transient cross-references. Otherwise calculating the reference
descriptions triggers the calculation of derived/transient features even
if no consumer has asked for the derived features to be calculated.

By doing this, we also keep it in closer to
LazyLinkingResource.resolveLazyCrossReference, which also skips this
types of EReferences.


gregdyke:
I'm convinced this is the correct thing to do. What I'm less sure is that the correctness of invalidations in ASMD doesn't currently rely on indexing transient/derived references

PR #399, opened 2021-06-08
Keep Sequence Number Suffix after call to addFragmentMapping

rubenporras:
Keep the sequence number suffix of the last iteration after the call to
addFragmentMapping.

If the counter starts each time at one, the while loop will increment
the counter by one and check the fragmentToObjectMap for existence of
the newly generated key until it reaches the last computed index,
always. This algorithm of O(n^2) complexity is a problem since the
existence check on fragmentToObjectMap is relatively expensive. Since we
know that all the previous indexes have already been given, we can start
at the last free index.

In AbstractExportedNameProvider.getFullyQualifiedName(EObject) a small
code cleanup has been performed.

### DIFF
--- a/com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/modelinference/impl/InferenceContainerImplCustom.java
+++ b/com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/modelinference/impl/InferenceContainerImplCustom.java
@@ -12,9 +12,11 @@
 package com.avaloq.tools.ddk.xtext.modelinference.impl;
 
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.log4j.Logger;
 import org.eclipse.emf.common.notify.Notification;
@@ -40,6 +42,7 @@ public class InferenceContainerImplCustom extends InferenceContainerImpl {
   private final BiMap<String, EObject> fragmentToObjectMap = HashBiMap.create();
   private final Set<String> fragmentSet = Collections.unmodifiableSet(fragmentToObjectMap.keySet());
   private final Map<EObject, String> objectToFragmentMap = fragmentToObjectMap.inverse();
+  private Map<String, AtomicInteger> collisionFragmentIndexMap;
 
   private IInferredElementFragmentProvider fragmentProvider;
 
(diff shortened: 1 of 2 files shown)


com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/modelinference/impl/InferenceContainerImplCustom.java
```diff
@@ -91,10 +92,10 @@ private void addFragmentMapping(final EObject object) {
       fragmentSegment = Integer.toString(getContents().size() - 1);
     }
     String key = fragmentSegment;
-    int i = 1;
+
     while (fragmentToObjectMap.putIfAbsent(key, object) != null) {
       // append sequence number suffix in case fragment is already assigned (collision)
-      key = fragmentSegment + '.' + i++;
+      key = fragmentSegment + '.' + collisionFragmentIndex++;
```
gregdyke, line 98: shouldn't these i sequence number suffixes be different for different fragmentSegment values? Is that needed for fragment stability?

gregdyke (changes requested):

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/modelinference/impl/InferenceContainerImplCustom.java
rubenporras: Hi Greg,

yes, I could do that by keeping duplites fragments in a map from fragment to sequence, but I think it is not needed, with an increasing number, even if it ocunts over all fragments, the numbers are stable as well, as the features are loaded from the binary model in the same order as they are in the inference container.

So I think keeping track of the duplicated fragments could render nicer URIs, but I do not think it is worth it, but if you think otherwise I can do it

gregdyke: I think I have a very limited understanding of the topic, so could be completely wrong/misguided.

If we have a model with names

- foo
- foo
- bar
- bar
- bar

that becomes


- foo.1 (previously foo.1) (or maybe both of these are foo?)
- foo.2 (previously foo.2)
- bar.3 (previously bar.1) (or maybe both of these are bar?)
- bar.4 (previously bar.2)
- bar.5 (previously bar.3)

Then if we remove a foo, the bar fragments also get renumbered, which I think impacts fragment stability and fine grained invalidation?

rubenporras: Yes, the fragments you wrote also match my understanding. Before we had bar.1, bar.2, bar.3, now we would have bar.3, bar.4 and bar.5.

And yes, I think you are right as well, if we remove foo.1, before only foo.2 would shift, now all bar would be shifted as well. I think this is not very bad, because if the objects are exported, the fragments must be unique to begin, so we should never fall into the collision case, otherwise we have a problem already with the previous code. As a matter of fact, the colliding fragments belong to implicit local variables, which are all named, for example "l_res", and those are not exported.

In any case, I see your point, it is not very hard to keep the IDs starting by 1 for each colliding fragment, as it was before, and still improve performance, so even though I think your scenario should not be a new problem, we do not need to make it worse, I will submit a new version.

gregdyke (approved):

PR #400, opened 2021-06-10
Strip bundle versions and disable preferXtendStubs

rubenporras:
Strip bundle versions as we do not want to have minimal versions in our
MANIFESTs for xtext and xbase.

Disable preferXtendStubs, as we prefer Java for stub files to be
implemented manually.


gregdyke (commented):
Probably these can be moved to defaults in our ddk fragments in a future change? But otherwise they are good there

gregdyke (approved):

PR #406, opened 2021-07-06
Fix388

rubenporras:


com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/AbstractCachingResourceDescriptionManager.java
```diff
@@ -276,7 +271,7 @@ public boolean isAffected(final Collection<Delta> deltas, final IResourceDescrip
         if (!references.contains(uri) && filter.test(uri)) {
           final List<IContainer> containers = getContainerManager().getVisibleContainers(res, context);
           for (IContainer container : containers) {
-            if (container.getResourceDescription(uri) != null) {
+            if (container.getResourceDescription(uri) != null && containersState.getContainerHandle(uri).equals(containerHandle)) {
```
andrewL-avlq, line 274: The added condition does not appear to depend on 'container' - does it? should it? or could it be moved out of the loop?

PR #418, opened 2021-07-22
Cache classes in getService

rubenporras:
Cache classes in getService, as calling getInstance() on each injector
of each resource is expensive, and we know, for the same file extension,
always the same class will be returned.

### DIFF
--- a/com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/linking/LazyLinkingResource2.java
+++ b/com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/linking/LazyLinkingResource2.java
@@ -11,6 +11,8 @@
 package com.avaloq.tools.ddk.xtext.linking;
 
 import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Set;
 
 import org.apache.log4j.Logger;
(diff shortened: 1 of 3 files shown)


com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/linking/LazyLinkingResource2.java
```diff
@@ -184,7 +188,22 @@ public synchronized EObject getEObject(final String uriFragment) {
...
-    return injector.getInstance(key);
+    IResourceSetServiceCache serviceCache = ServiceCacheAdapter.getServiceCacheAdapter(getResourceSet());
+    String fileExtension = getURI().fileExtension();
+    Map<Class<?>, Object> map = serviceCache.get(fileExtension);
+    if (map == null) {
+      map = new HashMap<>();
+      serviceCache.put(fileExtension, map);
+    }
+    @SuppressWarnings("unchecked")
+    T clazz = (T) map.get(key);
```
andrewL-avlq, line 199: 'clazz' is usually used as the name for variables of type Class<>, but here the variable is of type T. Maybe 'service' would be a better name?

rubenporras: You are right, I have changed it.

andrewL-avlq (approved):

PR #419, opened 2021-07-27
Add context sensitive api to UniquenessValidValidationHelper.

andrewL-avlq:
Added checkDuplicates api to UniquenessValidValidationHelper that takes
a context so that only messages for the context's resource are returned.
This api prevents exceptions being thrown when there are duplicates
outside the current resource and we try to attach messages for them.

### DIFF
--- a/com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/validation/UniquenessValidValidationHelper.java
+++ b/com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/validation/UniquenessValidValidationHelper.java
@@ -10,9 +10,11 @@
  *******************************************************************************/
 package com.avaloq.tools.ddk.xtext.validation;
 
+import java.util.Objects;
 import java.util.Set;
 
 import org.eclipse.emf.ecore.EObject;
+import org.eclipse.emf.ecore.resource.Resource;
 import org.eclipse.osgi.util.NLS;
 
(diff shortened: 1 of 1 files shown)


com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/validation/UniquenessValidValidationHelper.java
```diff
@@ -58,4 +59,34 @@ public UniquenessValidValidationHelper(final INameFunction nameFunction) {
...
+   *
+   * @param context
+   *          the context containing the possible duplicates.
+   * @param possiblyDuplicateObjects
+   *          a set of elements of type T, into which one must look for duplicate objects
+   * @param message
+   *          the duplicate error message, in which {0} is the type of T and {1} is the duplicate representation
+   * @return a compound message (a set of simple messages)
+   */
+  public CompoundValidMessage<T> checkDuplicates(final EObject context, final Iterable<T> possiblyDuplicateObjects, final String message) {
```
enzosanmiguel, line 73: Perhaps we should also do a null check on the context, or add in the javadoc if context must not be null

andrewL-avlq: I've added a null check and a null-safe equals call.

enzosanmiguel (approved):

PR #420, opened 2021-07-28
Add logging information for null EObject

KrisLimbo:
Added checking for null context EObject.
If context EObject is not null, then accept the issue.
Otherwise, log the context information for debugging.


andrewL-avlq:
Looks ok to me. Anyone object to me approving it?

enzosanmiguel (approved):

PR #422, opened 2021-08-03
Modified checking for null context EObject in check file.

KrisLimbo:
If issue expression is equal to the implicit/input variable,
no null checking for context EObject is performed.
Otherwise, perform null checking and log context information
for debugging null EObject context object.

### DIFF
--- a/com.avaloq.tools.ddk.check.core/src/com/avaloq/tools/ddk/check/generator/CheckCompiler.java
+++ b/com.avaloq.tools.ddk.check.core/src/com/avaloq/tools/ddk/check/generator/CheckCompiler.java
@@ -264,14 +264,21 @@ public class CheckCompiler extends XbaseCompiler {
       b.newLine();
     }
 
-    // checking for null context EObject
-    b.append("if (");
+    boolean issueExpressionEqualsImplicitVariable = true;
     if (markerObject != null) {
-      internalToConvertedExpression(markerObject, b, getLightweightType(eObjectType));
-    } else {
-      b.append(getContextImplicitVariableName(expr));
+      issueExpressionEqualsImplicitVariable = markerObject.toString().equalsIgnoreCase(getContextImplicitVariableName(expr));
     }
-    b.append(" != null) {").increaseIndentation().newLine();
+
+    if (!issueExpressionEqualsImplicitVariable) {
+      // checking for null context EObject
+      b.append("if (");
+      if (markerObject != null) {
+        internalToConvertedExpression(markerObject, b, getLightweightType(eObjectType));
+      } else {
+        b.append(getContextImplicitVariableName(expr));
+      }
+      b.append(" != null) {").increaseIndentation().newLine();
+    }
 
     // acceptor
     b.append("// Issue diagnostic").newLine();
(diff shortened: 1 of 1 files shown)


andrewL-avlq (approved):

com.avaloq.tools.ddk.check.core/src/com/avaloq/tools/ddk/check/generator/CheckCompiler.java
```diff
@@ -264,14 +264,21 @@ protected void _toJavaStatement(final XIssueExpression expr, final ITreeAppendab
...
     }
 
-    // checking for null context EObject
-    b.append("if (");
+    boolean issueExpressionEqualsImplicitVariable = true;
     if (markerObject != null) {
-      internalToConvertedExpression(markerObject, b, getLightweightType(eObjectType));
-    } else {
-      b.append(getContextImplicitVariableName(expr));
+      issueExpressionEqualsImplicitVariable = markerObject.toString().equalsIgnoreCase(getContextImplicitVariableName(expr));
```
gregdyke, line 269: I don't think equalsIgnoreCase is what we want here (but can fix this in a followup)

com.avaloq.tools.ddk.check.core/src/com/avaloq/tools/ddk/check/generator/CheckCompiler.java
```diff
@@ -264,14 +264,21 @@ protected void _toJavaStatement(final XIssueExpression expr, final ITreeAppendab
...
+    }
+
+    if (!issueExpressionEqualsImplicitVariable) {
+      // checking for null context EObject
+      b.append("if (");
+      if (markerObject != null) {
+        internalToConvertedExpression(markerObject, b, getLightweightType(eObjectType));
+      } else {
+        b.append(getContextImplicitVariableName(expr));
```
gregdyke, line 278: this no longer needs to be here?

gregdyke: specifically the else, as we don't need to test if contextImplicitVariable is null

gregdyke: I take it back, we don't need the whole if at line 275, based on what we already have at line 268

KrisLimbo: Thank you for the comment. It is redundant.

com.avaloq.tools.ddk.check.core/src/com/avaloq/tools/ddk/check/generator/CheckCompiler.java
KrisLimbo: Done.

com.avaloq.tools.ddk.check.core/src/com/avaloq/tools/ddk/check/generator/CheckCompiler.java
KrisLimbo: Done.

PR #439, opened 2021-09-17
Do not generate code when guard is false in ResourceDescriptionStrategyGenerator

markdomeng:
this prevents the generator from generating a dead end code 'if (false) {'


com.avaloq.tools.ddk.xtext.export/src/com/avaloq/tools/ddk/xtext/export/generator/ResourceDescriptionStrategyGenerator.xtend
```diff
@@ -82,7 +82,7 @@ class ResourceDescriptionStrategyGenerator {
                 «javaContributorComment(c.location)»
                 @Override
                 public Boolean case«c.type.name»(final «c.type.instanceClassName()» obj) {
-                  «IF c.guard !== null»
+                  «IF c.guard !== null && !c.guard.javaExpression(ctx.clone('obj', c.type)).equalsIgnoreCase("false")»
```
andrewL-avlq, line 85: Is it too much of a mess to pull c.guard.javaExpression(ctx.clone('obj', c.type)) into a common variable so that it is more obvious that you are checking the same form of the guard expression as used in the generated code (line 87 below)?

markdomeng: done

rubenporras (approved):

PR #441, opened 2021-09-23
Fix a bug in ResourceDescriptionStrategyGenerator 

markdomeng:
from commit #439, where it supposed to not generate the code anymore when the guard is null

### DIFF
--- a/com.avaloq.tools.ddk.xtext.export/src/com/avaloq/tools/ddk/xtext/export/generator/ResourceDescriptionStrategyGenerator.xtend
+++ b/com.avaloq.tools.ddk.xtext.export/src/com/avaloq/tools/ddk/xtext/export/generator/ResourceDescriptionStrategyGenerator.xtend
@@ -88,7 +88,7 @@ class ResourceDescriptionStrategyGenerator {
                     if («guard») {
                       «generateCaseBody(c, ctx, genModelUtil)»
                     }
-                  «ELSE»
+                  «ELSEIF c.guard === null»
                     «generateCaseBody(c, ctx, genModelUtil)»
                   «ENDIF»
 


com.avaloq.tools.ddk.xtext.export/src/com/avaloq/tools/ddk/xtext/export/generator/ResourceDescriptionStrategyGenerator.xtend
```diff
@@ -83,7 +83,8 @@ class ResourceDescriptionStrategyGenerator {
                 @Override
                 public Boolean case«c.type.name»(final «c.type.instanceClassName()» obj) {
                   «val guard = c.guard.javaExpression(ctx.clone('obj', c.type))»
-                  «IF c.guard !== null && !guard.equalsIgnoreCase("false")»
+                  «IF guard.equalsIgnoreCase("false")»
+                  «ELSEIF c.guard !== null»
```
andrewL-avlq, line 87: Having an empty conditional block looks wrong to me. How about replacing the «ELSE» on line 91 with «ELSEIF c.guard == null» ?

com.avaloq.tools.ddk.xtext.export/src/com/avaloq/tools/ddk/xtext/export/generator/ResourceDescriptionStrategyGenerator.xtend
```diff
@@ -88,7 +88,7 @@ class ResourceDescriptionStrategyGenerator {
                     if («guard») {
```
rubenporras, line 88: Should the block not be?

                  «IF c.guard === null»
                    «generateCaseBody(c, ctx, genModelUtil)»
                  «ELSIF !guard.equalsIgnoreCase("false")»
                    «javaContributorComment(c.guard.location)»
                    if («guard») {
                      «generateCaseBody(c, ctx, genModelUtil)»
                    }
                  «ENDIF»

this way the is no need to check twice if c.guard is null or not

andrewL-avlq (approved):

PR #466, opened 2021-11-11
Upgrade spotbugs

rubenporras:


rubenporras:
This PR upgrade Spotbugs, enables it by removing the "skip" flag in the pom.xml, and fixes the warnings.

ava-fred (approved):

PR #476, opened 2021-11-19
Fix broken unit tests in Eclipse

ava-fred:
The Eclipse project config change in PR 474 (commit
de70e081725e08007bcb1fe6491a971e13c9108e) had the consequence that unit
tests failed in Eclipse.
This change fixes the problem by changing the project config back, but
keeping the ability to at least filter the errors on test sources from
the Problems view in Eclipse.
It also includes a minor fix for an error in Eclipse and the addition of lsp4e to the dslsdk target.

### DIFF
--- a/com.avaloq.tools.ddk.check.core/metamodel/com/avaloq/tools/ddk/check/Check.xtext
+++ b/com.avaloq.tools.ddk.check.core/metamodel/com/avaloq/tools/ddk/check/Check.xtext
@@ -164,5 +164,6 @@ FeatureCallID:
 enum SeverityKind:
   error='error' | warning='warning' | info='info' | ignore='ignore';
 
-enum TriggerKind:  // we avoid using CheckKind to simplify writing expressions...
+// we avoid using CheckKind to simplify writing expressions...
+enum TriggerKind:
   fast='live' | normal='onSave' | expensive='onDemand';
--- a/ddk-target/ddk.target
+++ b/ddk-target/ddk.target
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <?pde version="3.8"?>
-<target name="DDK Target" sequenceNumber="12">
+<target name="DDK Target" sequenceNumber="13">
 <locations>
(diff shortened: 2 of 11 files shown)


com.avaloq.tools.ddk.check.core/metamodel/com/avaloq/tools/ddk/check/Check.xtext
```diff
@@ -164,5 +164,6 @@ FeatureCallID:
 enum SeverityKind:
   error='error' | warning='warning' | info='info' | ignore='ignore';
 
-enum TriggerKind:  // we avoid using CheckKind to simplify writing expressions...
+// we avoid using CheckKind to simplify writing expressions...
+enum TriggerKind:
```
rubenporras, line 168: fine for me, just as information, this bug has been fixed in the upcomming Xtext 2.26

ddk-target/ddk.target
```diff
@@ -35,5 +35,9 @@
 <unit id="org.eclipse.xtext.sdk.feature.group" version="0.0.0"/>
 <repository location="http://download.eclipse.org/modeling/tmf/xtext/updates/releases/2.25.0/"/>
 </location>
+<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
+  <repository location="https://download.eclipse.org/lsp4e/releases/0.16.0/"/>
+  <unit id="org.eclipse.lsp4e" version="0.13.3.202008291637"/>
+</location>
```
rubenporras, line 41: why do we need lsp4e on the ddk target?

ava-fred: It fixes an error in the c.a.t.ddk.xtext.ide project in a DDK-only workspace

rubenporras: I looked at the manifest of that project, and I think adding org.eclipse.lsp4j ('j', not 'e' at the end) would be enough. Could you try that?

ava-fred: done, works in my workspace

rubenporras (approved):

PR #494, opened 2021-11-30
Update to java11

rubenporras:


ava-fred (changes requested):
This has to wait until we have changed all our startup scripts to use Java11 and we have rebased to ICE 15.1

rubenporras:
@ava-fred , I agree, I wanted to be sure that you do not forget I have already done this change.

ava-fred (approved):

PR #496, opened 2021-11-30
remove setting of genericIdeSuffix

hasu:
Don't unnecessarily assign the suffix in the forceDisableIdeProject case
as it apparently does not get used in that case.


hasu:
If genericIdeSuffix is only used for naming the IDE project, as it looks like, then I guess its value does not matter when that project is not being generated, and the default assignment can be removed.

rubenporras (approved):

PR #505, opened 2021-12-02
Fix a bug in ResourceDescription2 where the type is not exported if only the supertype is exported

markdomeng:
the call to Eclass.isSupertypeOf(someclass) is misleading, as we want to check if the type is being exported if its supertype is exported


rubenporras:
Hi @markdomeng ,

how did we discover this bug? Do we have an example of how it manifested itself?

rubenporras (changes requested):
It looks like this is not the right approach, according to the meeting yesterday

rubenporras (commented):
Hi Mark,

this PR looks good to me but I will wait for the approval until we release the current master branch, which will happen next week

PR #526, opened 2022-01-17
Do not create a DelegatingResourceDescriptionsData without a delegate

rubenporras:
Do not create a DelegatingResourceDescriptionsData without a delegate as
this will crash with an NPE whenever a method of the
DelegatingResourceDescriptionsData is called (all of them used the
delegate). Instead, create a NullResourceDescriptionsData.

This prevents crashing on the Language Server when a context is
activated and the document highlight is run.


ava-fred (approved):
Change seems risk free - I'm leaving it up to you whether you want to merge now or wait until after release

PR #528, opened 2022-01-18
Fix crashes in the context of a Language Server

rubenporras:
* As in the previous change, do not create a ResourceSetAware without a
delegate as this will crash with an NPE whenever any method is called
(all of them used the delegate). Instead, create a
NullResourceDescriptionsData.

* Avoid processing the resource description currently being added to the
index in ResourceDescriptionsUtil.findReferencesToResources. The class
used by the language server for those descriptors
(Indexer$ResolvedResourceDescription) logs an error if getImportedNames
or getReferenceDescriptions is called. The reason of the Xtext project
for this exception is to detect errors
(https://github.com/eclipse/xtext-core/pull/1807#discussion_r786297617):
"It's noteworthy that these descriptions have been created for resources
that have not been fully processed. If they remain in the index after
the build, subsequent changes cannot correctly detect the affected
resources and the index state will gradually erode and become different
from the actual project contents on disk.
The ResolvedDescriptions should be local to a single build run and never
survive that single build cycle. Within the build loop, there imported
names and reference descriptions should never be necessary."

* Inline with this reasoning, we bring
FixedCopiedResourceDescription closer to the Xtext class
CopiedResourceDescription and log the exceptional case whenever
possible.

### DIFF
--- a/com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/FixedCopiedResourceDescription.java
+++ b/com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/FixedCopiedResourceDescription.java
@@ -15,10 +15,12 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import org.apache.log4j.Logger;
 import org.eclipse.emf.common.util.URI;
 import org.eclipse.emf.ecore.EClass;
 import org.eclipse.emf.ecore.InternalEObject;
 import org.eclipse.emf.ecore.util.EcoreUtil;
+import org.eclipse.xtext.builder.clustering.CopiedResourceDescription;
 import org.eclipse.xtext.naming.QualifiedName;
 import org.eclipse.xtext.resource.EObjectDescription;
 import org.eclipse.xtext.resource.IEObjectDescription;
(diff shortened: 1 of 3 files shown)


com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/extensions/ResourceDescriptionsUtil.java
```diff
@@ -85,23 +86,25 @@ private ResourceDescriptionsUtil() {}
       }
     }
 
-    return Iterables.filter(descriptions.getAllResourceDescriptions(), input -> {
-      if (matchNames) {
-        for (QualifiedName name : input.getImportedNames()) {
-          if (exportedNames.contains(name.toLowerCase())) { // NOPMD
-            return true;
+    return Iterables.filter(//
+        Iterables.filter(descriptions.getAllResourceDescriptions(), input -> !targetResources.contains(input)) //
```
rubenporras, line 90: besides formatting, the main change in this class is to add the condition 
input -> !targetResources.contains(input)

com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/FixedCopiedResourceDescription.java
```diff
@@ -30,41 +32,45 @@
...
-        } else if (from instanceof IDetachableDescription) {
-          return ((IDetachableDescription<IEObjectDescription>) from).detach();
-        }
-        InternalEObject result = (InternalEObject) EcoreUtil.create(from.getEClass());
-        result.eSetProxyURI(from.getEObjectURI());
-        ImmutableMap.Builder<String, String> userData = ImmutableMap.builder();
-        for (final String key : from.getUserDataKeys()) {
-          userData.put(key, from.getUserData(key));
-        }
-        return EObjectDescription.create(from.getName(), result, userData.build());
```
rubenporras, line 66: the diff looks big, but I just turned the block into a lambda expression

andrewL-avlq (approved):

PR #539, opened 2022-02-03
Migrate away from legacy Xtext generator

rubenporras:
The old CustomClassEcoreGenerator and related classes have been removed.

The plugin-gen.xml sources changed the order of it entries and some
whitespace characters, since these changes do not have a functional
effect, the plugin.xml files have not been modified to avoid additional
noise when comparing them.


com.avaloq.tools.ddk.check.ui/plugin.xml_gen
```diff
@@ -1,215 +1,220 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
andrewL-avlq, line 1: Does any of this need merging with plugin.xml?

com.avaloq.tools.ddk.workflow/src/com/avaloq/tools/ddk/workflow/GenerateExport.mwe2
```diff
@@ -16,103 +16,13 @@ import org.eclipse.xtext.ui.generator.*
...
-
-            // provides a compare view
-            fragment = compare.CompareFragment {
-                fileExtensions = file.extensions
-            }
-
-        }
-    }
+@com.avaloq.tools.ddk.workflow.CommonXbase auto-inject {
+    languageName = "CheckCfg"
```
andrewL-avlq, line 25: languageName should be "Export" here no?

rubenporras: Right, this mwe2 file was wrongly, it does not use the CommonXbase. I have modified it now and left it standalone.

com.avaloq.tools.ddk.check.ui/plugin.xml_gen
rubenporras: I think it is better not to do that, so I have just added a comment in the commit.

com.avaloq.tools.ddk.check.core/META-INF/MANIFEST.MF
```diff
@@ -47,5 +47,6 @@ Export-Package: com.avaloq.tools.ddk.check,
  com.avaloq.tools.ddk.check.util,
  com.avaloq.tools.ddk.check.validation
 Import-Package: org.apache.logging.log4j,
- org.eclipse.jdt.internal.ui.text
+ org.eclipse.jdt.internal.ui.text,
+ org.apache.log4j
```
andrewL-avlq, line 51: Shouldn't we be using the new loggers everywhere now?

rubenporras: The Xtext generator puts those back. I can remove them by hand if you like, but regenerating will put them back. Do you think it is worth it?

com.avaloq.tools.ddk.xtext.expression/src-gen/com/avaloq/tools/ddk/xtext/expression/expression/BooleanLiteral.java
```diff
@@ -21,26 +21,26 @@
...
-   * <!-- begin-user-doc -->
-   * <!-- end-user-doc -->
-   * @return the value of the '<em>Val</em>' attribute.
-   * @see #setVal(String)
-   * @see com.avaloq.tools.ddk.xtext.expression.expression.ExpressionPackage#getBooleanLiteral_Val()
-   * @model
-   * @generated
-   */
-  String getVal();
+	/**
```
andrewL-avlq: This (and several others in this plugin) appears to just contain whitespace changes that replace 2-spaces with 1-tab. Has the configuration for something changed?

rubenporras: My findings are the following:
- In ASMD and other plugins we have tabs, so I think this plug-in was not generated in a very long time, and the code generator behaviour has changed, why I do not know.
- The generator should pick the project formatter settings https://www.eclipse.org/forums/index.php/t/1065524/, I checked our project settings and we have org.eclipse.jdt.core.formatter.tabulation.char=space
org.eclipse.jdt.core.formatter.tabulation.size=2, org.eclipse.jdt.core.formatter.use_tabs_only_for_leading_indentations=true. I have set the later to false and restarted eclipse, just in case, and it did not do any difference
- The generator uses internally a JMerge https://www.eclipse.org/modeling/emf/docs/2.x/whatsnew/merge2.3.html, which offers the possibility to be configured, but I could not figure out how. I tried to specify in the genmodel the property codeFormatting="true" but it did not do any difference.
- The default for JMerge should be actually spaces, the file /templates/emf-merge.xml in the org.eclipse.emf.codegen.ecore plug-in has the following

<merge:options 
  indent="  "  
  braceStyle="matching"  
  redirect="Gen"
  block="\s*@\s*generated\s*NOT\s*(?:\n\r?|\r\n?)"
  noImport="\s*//\s*import\s+([\w.*]*)\s*;\s*(?:\n\r?|\r\n?)"
  xmlns:merge="http://www.eclipse.org/org/eclipse/emf/codegen/jmerge/Options">


- Finally, I got it to generate spaces by overriding the creation of the JControlModel, I will show you this later, but it does not look as if this is supposed to be customized like that.

andrewL-avlq (dismissed):

andrewL-avlq (approved):

PR #605, opened 2022-06-07
Added logging to catch null resourceSet when some resources are saved. 

KrisLimbo:


rubenporras (changes requested):
Could we put the //$NON-NLS-1$ in a different PR, so that the logging can be reverted independent of it?

com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
```diff
@@ -694,29 +696,31 @@ protected Resource addResource(final Resource resource, final ResourceSet resour
    *          build data, must not be {@code null}
    */
   protected void storeBinaryResource(final Resource resource, final BuildData buildData) {
-    if (isBinaryModelStorageAvailable && resource instanceof StorageAwareResource && ((StorageAwareResource) resource).getResourceStorageFacade() != null
+    if (resource.getResourceSet() == null) {
```
rubenporras, line 699: we should do this validation after if (isBinaryModelStorageAvailable && resource instanceof StorageAwareResource && ((StorageAwareResource) resource).getResourceStorageFacade() != null
otherwise we could complain about resources we will never save, and that is not needed

com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
```diff
@@ -632,7 +632,9 @@ public boolean isCanceled() {
           if (resource instanceof XtextResource) {
             ((XtextResource) resource).getCache().clear(resource);
           }
-          storeBinaryResource(resource, buildData);
+          if (resource != null) {
```
rubenporras, line 635: I do not think this if is needed, because the first thing storeBinaryResource will do is to check resource instanceof ..., and that is only true if resource is not null

com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
```diff
@@ -694,29 +696,31 @@ protected Resource addResource(final Resource resource, final ResourceSet resour
    *          build data, must not be {@code null}
    */
   protected void storeBinaryResource(final Resource resource, final BuildData buildData) {
-    if (isBinaryModelStorageAvailable && resource instanceof StorageAwareResource && ((StorageAwareResource) resource).getResourceStorageFacade() != null
+    if (resource.getResourceSet() == null) {
+      LOGGER.info("null resourceSet found for " + resource.getURI()); //$NON-NLS-1$
```
rubenporras, line 700: no resourceSet found for... reads better

rubenporras (changes requested):

com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
KrisLimbo: You're right.

com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
KrisLimbo: I am not a fan of this particular change either.
However, previous verify jobs (SpotBugs) complain about it because it is initialized (line 523) as null and it is possible for addResource (line 536) to return null as well.
Do we have an alternative (like suppress checking for this particular case)?

com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
KrisLimbo: Ok.

com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
KrisLimbo: Uploaded another change for this one. Thank you.

rubenporras (approved):
thanks

PR #613, opened 2022-06-28
catch stack overflows when loading from storage

hasu:
That is, catch and log StackOverflowErrors when loading resources from
storage, and fall back to recreating the resource from source.

### DIFF
--- a/com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
+++ b/com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
@@ -92,6 +92,7 @@ import com.avaloq.tools.ddk.xtext.scoping.ImplicitReferencesAdapter;
 import com.avaloq.tools.ddk.xtext.tracing.ITraceSet;
 import com.avaloq.tools.ddk.xtext.tracing.ResourceValidationRuleSummaryEvent;
 import com.avaloq.tools.ddk.xtext.util.EmfResourceSetUtil;
+import com.avaloq.tools.ddk.xtext.util.ThrowableUtil;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableListMultimap;
 import com.google.common.collect.Iterables;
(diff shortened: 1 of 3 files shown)


com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/linking/LazyLinkingResource2.java
```diff
@@ -134,7 +136,64 @@ public void setLoading(final boolean loading) {
...
+  /**
+   * Modify the argument StackOverflowError so that it only retains
+   * the first and last {@link #STACK_TRACE_LIMIT} stack trace elements of {@code ex}
+   * if it has more than that.
+   *
+   * @param ex
+   *          the StackOverflowError; may get modified
+   */
+  @SuppressWarnings("nls")
+  private static void trimStackOverflowErrorStackTrace(final StackOverflowError ex) {
```
rubenporras, line 186: could we not refactor com.avaloq.tools.ddk.xtext.builder.MonitoredClusteringBuilderState.logStackOverflowErrorStackTrace(StackOverflowError, URI) and use it?
I would think that if we remove the URI parameter and do the logging outside as you do here, one could reuse the code. Shouldn't we do that?

rubenporras (changes requested):

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/linking/LazyLinkingResource2.java
hasu: OK, I've moved the common exception trimming code into a utility method.

com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
```diff
@@ -758,17 +758,12 @@ protected void awaitBinaryStorageExecutorQuiescence() {
    * Log the first and last 10 StackOverflowError Stack Trace.
    *
    * @param ex
-   *          the StackOverflowError
+   *          the StackOverflowError; may get modified
    * @param binding
    *          the object to be inserted into the message
    */
   protected void logStackOverflowErrorStackTrace(final StackOverflowError ex, final URI binding) {
```
rubenporras, line 765: I would remove the method and move the code to the exception handler, then it is clear that modifying the exception is not an issue because after LOGGER.warn it will go out of scope

hasu: Done.

rubenporras (approved):

PR #619, opened 2022-07-04
catch stack overflows when loading from storage

hasu:
That is, catch and log StackOverflowErrors when loading resources from
storage, and fall back to recreating the resource from source.


com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
```diff
@@ -749,24 +751,6 @@ protected void awaitBinaryStorageExecutorQuiescence() {
...
 
-  /**
-   * Log the first and last 10 StackOverflowError Stack Trace.
-   *
-   * @param ex
-   *          the StackOverflowError
-   * @param binding
-   *          the object to be inserted into the message
-   */
-  protected void logStackOverflowErrorStackTrace(final StackOverflowError ex, final URI binding) {
```
rubenporras, line 760: since this is a FP, to keep changes to the minimum, I would not delete this method, so you that you do not need to adapt subclasses.

rubenporras (approved):
You can submit as it is if you do not mind to refactor ASMD as well, but maybe you want keep the old method to avoid that

com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
hasu: Good idea, will keep the method. Also need to modify the PR to be against v9.4.x, I think.

PR #675, opened 2022-11-15
Mr QA: avoid SWTBot race condition to fix failing SWTBot tests

ava-fred:
There appears to be a race condition in SWTBot triggered by code in the DynamicViewMenu class, whereby a widget was disposed too early. This change re-implements DynamicViewMenu in a way that avoids this race condition.


rubenporras:
@andrewL-avlq , can I leave this to you? You know this code much better than me.

andrewL-avlq (approved):

PR #678, opened 2022-11-15
Delay testing resource set when storing binary resources.

andrewL-avlq:
Reduce the window where a resource could be unexpectedly removed from a resource set before the binary model is saved by delaying the check to occur in the async work along with the actual model persistence.

### DIFF
--- a/com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
+++ b/com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
@@ -122,6 +122,8 @@ public class MonitoredClusteringBuilderState extends ClusteringBuilderState
 
   private static final int COMMIT_WARN_WAIT_SEC = 30;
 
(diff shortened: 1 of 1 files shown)


com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
```diff
@@ -686,8 +686,8 @@ protected void storeBinaryResource(final Resource resource, final BuildData buil
     if (isBinaryModelStorageAvailable && resource instanceof StorageAwareResource && ((StorageAwareResource) resource).getResourceStorageFacade() != null
         && fileSystemAccess instanceof IFileSystemAccessExtension3) {
 
-      if (resource.getResourceSet() != null) {
-        CompletableFuture.runAsync(() -> {
+      CompletableFuture.runAsync(() -> {
+        if (resource.getResourceSet() != null) {
```
rubenporras, line 690: You need to extract this lambda to an own method for Checkstyle to accept the change, otherwise, it looks good.

rubenporras (approved):

PR #686, opened 2022-12-01
Update pom.xml

rubenporras:
Set the missing dependency for pmd-core, setting one for pmd-java is not enough to force the pmd maven plugin to use the desired version.


ava-fred (approved):
I have approved the changes, but "update pom.xml" is a really bad description for a PR. I can see that it updates pom.xml, but I have no idea why it does that.

rubenporras:
Hi @ava-fred , that is the commit ID, the PR description is different (I have adapted it by hand), maybe you have missed that?

rubenporras:
I have also adapted the commit message when squashing it.

PR #695, opened 2022-12-15
Ensure "Last Modified" metadata is constant for .checkbin files

joaodinissf:


com.avaloq.tools.ddk.check.core/src/com/avaloq/tools/ddk/check/resource/CheckBatchLinkableResourceStorageWritable.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipOutputStream;
+
+import org.eclipse.xtext.resource.persistence.StorageAwareResource;
+import org.eclipse.xtext.xbase.resource.BatchLinkableResource;
+import org.eclipse.xtext.xbase.resource.BatchLinkableResourceStorageWritable;
+
+
+@SuppressWarnings("restriction")
+public class CheckBatchLinkableResourceStorageWritable extends BatchLinkableResourceStorageWritable {
```
rubenporras, line 27: Could we comment in the javadoc what is special about our class compared to the base class?

rubenporras (commented):
All good appart from the small comment so that it is easy to spot the differences

rubenporras (approved):

PR #739, opened 2023-05-12
Add LSP setup to CheckCfg.

andrewL-avlq:
Adds fragment to CheckCfg workflow that generates the setup classes for LSP and workflow re-run.


com.avaloq.tools.ddk.checkcfg.core/META-INF/MANIFEST.MF
```diff
@@ -17,6 +17,7 @@ Require-Bundle: org.eclipse.xtext,
  org.eclipse.xtext.xbase;visibility:=reexport,
  org.eclipse.xtext.common.types,
  org.eclipse.jdt.core,
+ io.github.classgraph,
```
rubenporras, line 20: why do we need this extra dependency in the MANIFEST? There are no other changes than this one in the core plugin

rubenporras (dismissed):

rubenporras (approved):

PR #760, opened 2023-06-01
Upgrade to Xtext 2.31

rubenporras:


rubenporras:
I do not plan to merge it yet, but the PR would be ready

PR #770, opened 2023-06-16
Remove dependencies on GEF

rubenporras:


joaodinissf (changes requested):
Should we also get rid of all the matches of `.gef` in `AllTests DDK.launch` (8 total)?

joaodinissf (approved):

PR #775, opened 2023-06-22
Remove the Xtext version number from generated files

rubenporras:
to avoid polluting the git history on each Xtext upgrade


joaodinissf (commented):
Should the change also affect `CommonXbase.mwe2`, `GenerateExpression.mwe2`, and `GenerateScope.mwe2`?

joaodinissf (approved):

PR #783, opened 2023-06-27
Fix NPE

rubenporras:


joaodinissf (approved):
Your change should fix the NPE.

Would it make sense to consider fixing `readCompressedInt()` not to have these unintuitive side effects? For example, by storing resourceSet in a local variable, and restoring it after calling `super.readCompressedInt()`? That might incur a performance penalty, of course...

rubenporras:
> Your change should fix the NPE.
> 
> Would it make sense to consider fixing `readCompressedInt()` not to have these unintuitive side effects? For example, by storing resourceSet in a local variable, and restoring it after calling `super.readCompressedInt()`? That might incur a performance penalty, of course...

Yes, it would be good to do, but I am currently doing something else and do not want to expend much time into figuring out how to effectively remove the hack.

PR #828, opened 2023-11-09
Fix PMD Warning UnusedAssignment

rubenporras:


com.avaloq.tools.ddk.test.ui/src/com/avaloq/tools/ddk/test/ui/swtbot/CoreSwtbotTools.java
```diff
@@ -223,15 +223,13 @@ public static void captureScreenshot(final SWTBot bot, final String prefix, fina
...
     Assert.isNotNull(windowName, "windowName");
-    Boolean windowFound = false;
     try {
       final SWTBotShell shell = bot.shell(windowName);
       shell.isActive();
-      windowFound = true;
+      return true;
     } catch (WidgetNotFoundException exception) {
       throw new WrappedException("Error during searching for window", exception);
     }
```
joaodinissf, line 232: Do we not need a
```suggestion
    }
    return false;
```
here?

joaodinissf (changes requested):

com.avaloq.tools.ddk.test.ui/src/com/avaloq/tools/ddk/test/ui/swtbot/CoreSwtbotTools.java
rubenporras: I do not think so, that could would not be reachable, as either we return true or exit with the WrappedException.

joaodinissf (approved):

com.avaloq.tools.ddk.test.ui/src/com/avaloq/tools/ddk/test/ui/swtbot/CoreSwtbotTools.java
joaodinissf: Oops, you're right, of course. I misread the code when I left the comment.

PR #839, opened 2023-11-10
Remove Xtext dependency in cat.ddk.test.core plugin

joaodinissf:


joaodinissf:
This PR should not be merged until the downstream projects have been updated to reflect this refactoring.

PR #857, opened 2023-11-28
Do not use a live view of the ExtensionFactoryMap

rubenporras:
as this means that each time the extensions are accesed, the transform method will be applied. This has a high cost when the extension map is used for all the sources in the workspace. Since the registered extensions do not change once the system is setup, it is much more efficient to map the keys only once.


com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/ResourceServiceProviderUtil.java
```diff
@@ -12,13 +12,11 @@
 
 import java.util.Collection;
 import java.util.Set;
```
joaodinissf, line 14:
```suggestion
```

PR #858, opened 2023-11-28
Update class filter to reflect change in test runner.

andrewL-avlq:
The SwtBotRecordingTestRunner is now in a different package, so the SwtBotClassFilter needs to be updated otherwise no tests are found.


joaodinissf (approved):

joaodinissf (changes requested):
Sorry for withdrawing the approval--this looked right initially, but actually I think what we want to do is to use the new ClassFilter that was created for the new Xtext-specific test runner, `com.avaloq.tools.ddk.xtext.test.core/src/com/avaloq/tools/ddk/xtext/test/junit/runners/SwtBotXtextClassFilter.java`.

PR #882, opened 2024-01-17
Limit size of queue for storing binary models

ava-fred:
There is a queue of unprocessed binary models to be stored in MonitoredClusterinBuilderState. Previously, there was no size limit queue which means it could grow very large and cause memory issues.

This commit introduces this size limit. It is set to 15'000 by default and can be changed at run time by setting the system property com.avaloq.tools.ddk.xtext.builder.binaryStorageExecutor.queueCapacity.

The number of threads used to stored binary models is now also configurable by setting
com.avaloq.tools.ddk.xtext.builder.binaryStorageExecutor.parallelDegree.

### DIFF
--- a/com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
+++ b/com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
@@ -183,8 +183,8 @@ public class MonitoredClusteringBuilderState extends ClusteringBuilderState
     // @Format-On
   }
 
-  private int binaryStorageHighWaterMark = 0;
-  private LocalTime hwmTimeStamp = null;
+  private int binaryStorageHighWaterMark;
+  private LocalTime hwmTimeStamp;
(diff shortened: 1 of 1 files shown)


rubenporras (approved):

com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/MonitoredClusteringBuilderState.java
```diff
@@ -699,7 +712,12 @@ protected void storeBinaryResource(final Resource resource, final BuildData buil
         && fileSystemAccess instanceof IFileSystemAccessExtension3) {
 
       try {
-        CompletableFuture.runAsync(() -> doStoreBinaryResource(resource, buildData), binaryStorageExecutor);
+        int currentQueueSize = binaryStorageExecutor.getQueue().size();
+        if (currentQueueSize > binaryStorageHighWaterMark) {
```
rubenporras, line 716: what about recording this only if the queue is big enough? Let us say, > 1_000?

ava-fred: we could, but I don't see why we would want to

PR #888, opened 2024-01-30
Do not reject tasks for storing binary models

ava-fred:
In commit 4c15019, a upper limit for the number of unprocessed binary models was introduced. If this limit is reached, tasks are rejected if the default RejectedExecutionHandler of ThreadPoolExecutor is used. That is not what is intended, this commit changes the
RejectedExecutionHandler to CallerRunsPolicy.


rubenporras (approved):

ava-fred:
@rubenporras , that does not make sense to me. There shouldn't really be a release 12.5, the only reason there is is that we got the next version wrong when releasing.

PR #891, opened 2024-02-06
Remove obsolete parameter "ignoreTychoRepositories"

joaodinissf:
This option seems to have been removed in https://github.com/eclipse-tycho/tycho/commit/593a2467b7e5f39bb410f1e2ae4869576504873b (part of Tycho 0.14.0). See also https://github.com/eclipse-tycho/tycho/discussions/841.

Removing it should allow us to get rid of the warnings `Warning:  Parameter 'ignoreTychoRepositories' is unknown for plugin 'target-platform-configuration:3.0.5:target-platform (default-target-platform)'` in our Maven builds.

### DIFF
--- a/ddk-parent/pom.xml
+++ b/ddk-parent/pom.xml
@@ -195,7 +195,6 @@
               <classifier>${target.platform}</classifier>
             </artifact>
           </target>
-          <ignoreTychoRepositories>true</ignoreTychoRepositories>
           <environments>
             <environment>
               <os>${osgi.os}</os>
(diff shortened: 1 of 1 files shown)


rubenporras:
@ava-fred , could you take a look at this?

joaodinissf:
Added a proper description to the PR just now with the rationale for the removal. But please double-check if it makes sense to do so.

rubenporras (approved):

ava-fred:
Looks fine to me. In the unlikely case that there is a problem with it, I would expect the verify to fail

PR #906, opened 2024-03-21
Remove org.eclipse.xtext.generator consumers from DDK. (WIP)

KrisLimbo:
Removed org.eclipse.xtext.generator in required-bundle entry in MANIFEST.MF.

Removed org.eclipse.xtext.generator import in MWE2 workflow files.

Updated the following to use non-deprecated classes: CheckGenModelUtil, GenModelUtil2, ParserSemanticPredicatesUtil FormatGenerator, FormatUiModule

Removed .XPT files and fragments using deprecated classes in org.eclipse.xtext.generator.

Run DDK.MWE2.


KrisLimbo:
only testing

PR #911, opened 2024-03-27
Release 13.1.0.v20240327 1040 rel

rubenporras:


joaodinissf:
Could you do a fast-forward rebase instead of a merge, so this PR has only two commits? Rebasing on an up-to-date master from Eclipse followed by a force push to the branch should achieve this. I think that would be nicer.

gerdleon (approved):

joaodinissf (approved):
(Don't feel strongly enough about it to block merging, in case you want to go ahead as-is.)

rubenporras:
Will do a new PR

PR #926, opened 2024-04-22
Move find references filter (part 1).

andrewL-avlq:
Adds InferredModelReferenceFilter to ide plugin as first part of moving it from ui plugin.

### DIFF
--- a/com.avaloq.tools.ddk.xtext.ide/src/com/avaloq/tools/ddk/xtext/ide/findrefs/InferredModelReferenceFilter.java
+++ b/com.avaloq.tools.ddk.xtext.ide/src/com/avaloq/tools/ddk/xtext/ide/findrefs/InferredModelReferenceFilter.java
@@ -1,0 +1,50 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Avaloq Group AG and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Avaloq Group AG - initial API and implementation
+ *******************************************************************************/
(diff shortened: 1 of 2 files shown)


com.avaloq.tools.ddk.xtext.ide/src/com/avaloq/tools/ddk/xtext/ide/findrefs/InferredModelReferenceFilter.java
```diff
@@ -0,0 +1,54 @@
...
+   *
+   * @param elementURI
+   *          the URI to check
+   * @return true if {@code elementURI} is the URI of an inferred model element; false otherwise.
+   */
+  protected boolean isInferredModelElement(final URI elementURI) {
+    String fragment = elementURI.fragment();
+    if (fragment.length() > 1) {
+      char firstCharacter = fragment.charAt(1);
+      // Normally, URI fragments of non-inferred elements look like: /0/x/y/z/
```
rubenporras, line 45: We no longer have a custom FragmentProvider for WFD, it was used for the graphical editor and we have removed both. Have I missed something?

andrewL-avlq: This is just a copy of com.avaloq.tools.ddk.xtext.ui.editor.findrefs.InferredModelReferenceFilter - I have made no other changes. That said, even if the specific case in the comment is no longer present, shouldn't we still cope with anyone else that writes a custom uri fragment provider? (I think we should still change the comment though)

rubenporras: I could only find two in our codebase (IntfDef, CodeTab), and I am not sure, but I think they all look like /0/x/y/z, so we would not need the exception. Could you check?

rubenporras (approved):

PR #928, opened 2024-04-26
Remove rules which have been removed in PMD 7.

rubenporras:
The custom rule for detecting fall through in switch statements is deleted, as PMD has now a rule ImplicitSwitchFallThrough.

### DIFF
--- a/ddk-configuration/pmd/ruleset.xml
+++ b/ddk-configuration/pmd/ruleset.xml
@@ -24,21 +24,11 @@
     <exclude name="CloseResource"/>
     <exclude name="CompareObjectsWithEquals"/>
     <exclude name="ConstructorCallsOverridableMethod"/>
-    <exclude name="DataflowAnomalyAnalysis"/>
     <exclude name="DetachedTestCase"/>
     <exclude name="DontImportSun"/><!--Checkstyle-->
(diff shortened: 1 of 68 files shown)


joaodinissf (changes requested):
Looks like there's a few
`Warning: Empty Catch Block finds instances where an exception is caught, but nothing is done.`
that need fixing.

ddk-configuration/pmd/ruleset.xml
```diff
@@ -28,16 +28,6 @@
     <exclude name="DetachedTestCase"/>
     <exclude name="DontImportSun"/><!--Checkstyle-->
     <exclude name="DoNotTerminateVM"/>
-    <exclude name="EmptyCatchBlock"/><!--Checkstyle-->
```
joaodinissf, line 31: We probably should revert this line?

joaodinissf (approved):

joaodinissf (changes requested):
Change still required at https://github.com/dsldevkit/dsl-devkit/blob/2ac820ef84d04bed9c5c3976df038da2a95a95ca/.github/workflows/verify.yml#L16.

joaodinissf (approved):

PR #931, opened 2024-04-26
Upgrade PMD

rubenporras:


joaodinissf (changes requested):
This should be done together with your other open PR, I think.

PR #967, opened 2024-07-02
feat: add DispatchingCheckImpl check catalog base class

hasu:
That is, add DispatchingCheckImpl as an alternative base class for check catalog implementations, refactoring the common parts of DefaultCheckImpl into a common AbstractCheckImpl superclass.


com.avaloq.tools.ddk.check.runtime.core/src/com/avaloq/tools/ddk/check/runtime/issue/AbstractCheckImpl.java
```diff
@@ -0,0 +1,268 @@
...
+   * </ul>
+   *
+   * @param severity
+   *          the issue severity
+   * @return the numeric value representing a severity
+   */
+  protected static int toDiagnosticSeverity(final Severity severity) {
+    int diagnosticSeverity = -1;
+    switch (severity) {
+    case ERROR:
```
rubenporras, line 253: could we not use the new switch syntax to avoid the `break`s?

rubenporras (approved):

rubenporras (dismissed):

rubenporras (approved):

PR #976, opened 2024-07-26
feat: add AbstractDispatchingCheckImpl base class

hasu:
That is, create a variant of DispatchingCheckImpl as a slightly different check catalog implementation base class, one that assumes that the necessary validation state is passed around across 'validate' and context method calls as a reference to a DiagnosticCollector. Assume that instead of using a thread-local variable to track State, as is done by DispatchingCheckImpl.

To support implementation of the new CheckImpl base class do further refactoring of the existing ones.

A notable difference of AbstractDispatchingCheckImpl is that it no longer implements the ValidationMessageAcceptor interface, and it is instead the DiagnosticCollector that does.

### DIFF
--- a/com.avaloq.tools.ddk.check.runtime.core/src/com/avaloq/tools/ddk/check/runtime/issue/AbstractDispatchingCheckImpl.java
+++ b/com.avaloq.tools.ddk.check.runtime.core/src/com/avaloq/tools/ddk/check/runtime/issue/AbstractDispatchingCheckImpl.java
@@ -17,7 +17,10 @@ import java.util.concurrent.ConcurrentHashMap;
 import org.eclipse.emf.common.util.DiagnosticChain;
 import org.eclipse.emf.ecore.EClass;
 import org.eclipse.emf.ecore.EObject;
+import org.eclipse.xtext.validation.AbstractValidationMessageAcceptor;
 import org.eclipse.xtext.validation.CheckMode;
+import org.eclipse.xtext.validation.CheckType;
+import org.eclipse.xtext.validation.ValidationMessageAcceptor;
 
 import com.avaloq.tools.ddk.xtext.tracing.ITraceSet;
 import com.avaloq.tools.ddk.xtext.tracing.ResourceValidationRuleSummaryEvent;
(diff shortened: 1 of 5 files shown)


com.avaloq.tools.ddk.check.runtime.core/src/com/avaloq/tools/ddk/check/runtime/issue/AbstractStatefulCheckImpl.java
```diff
@@ -0,0 +1,96 @@
...
+  public ValidationMessageAcceptor getMessageAcceptor() {
+    return messageAcceptor;
+  }
+
+  // ////////////////////////////////////////////////
+  // Copied and adapted from AbstractDeclarativeValidator below
+  // ////////////////////////////////////////////////
+
+  /**
+   * The holding the current state for a validation method being executed.
```
joaodinissf, line 51: There was a missing word in the original code--maybe we can take this opportunity to fix it?
```suggestion
   * The class holding the current state for a validation method being executed.
```

com.avaloq.tools.ddk.check.runtime.core/src/com/avaloq/tools/ddk/check/runtime/issue/AbstractDispatchingCheckImpl.java
```diff
@@ -0,0 +1,186 @@
...
+   * @param e
+   *          the exception that was raised
+   */
+  protected void handleContextMethodFailure(final String contextName, final EObject object, final Exception e) {
+    disabledMethodTracker.setDisabled(contextName);
+    logCheckMethodFailure(getQualifiedCatalogName() + contextName, object, e);
+  }
+
+  /**
+   * The holding the current state for a validation method being executed.
```
joaodinissf, line 146:
```suggestion
   * The class holding the current state for a validation method being executed.
```

com.avaloq.tools.ddk.check.runtime.core/src/com/avaloq/tools/ddk/check/runtime/issue/AbstractDispatchingCheckImpl.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    @Override
+    public void setCurrentCheckType(final CheckType checkType) {
+      currentCheckType = checkType;
+    }
+  }
+
+  /**
+   * An implementation of {@link DiagnosticCollector} that does nothing.
```
joaodinissf, line 177: Why do we need this? Can we maybe add a brief comment explaining it?

joaodinissf (changes requested):

joaodinissf (approved):

PR #1002, opened 2024-09-24
Refactor to allow reducing dependencies to ddk.xtext.builder

rubenporras:


com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/layered/IXtextTargetPlatform.java
```diff
@@ -49,7 +47,6 @@
  * encapsulated index or the issue store are accessed, the platform must be opened through {@link #open(boolean, IProgressMonitor)}. Once open, all operations
  * are allowed. Once it is no longer in use, a platform must be closed through {@link #close(IProgressMonitor)}.
  */
-@ImplementedBy(DefaultXtextTargetPlatform.class)
```
rubenporras, line 52: @andrewL-avlq , will I need to bind this now in ASMD or do we have a more specific XtextTargetPlatform?

andrewL-avlq: I think we do not inject IXtextTargetPlatforms directly, but instead bind the IXtextTargetPlatformFactory and inject/use that instead.

andrewL-avlq (approved):

PR #1031, opened 2024-11-20
upgrade to Xtext 2.37

rubenporras:


ddk-target/ddk-antlr.target
joaodinissf, line 1: Should the sequence number also be bumped here?

joaodinissf (changes requested):

joaodinissf (approved):

PR #1047, opened 2024-12-03
Reduce duplicated access to a map element

rubenporras:


com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/XtextGMFLazyLinkingResource.java
```diff
@@ -88,7 +88,7 @@ protected void loadFromFullContents(final InputStream input, final String encodi
    */
   private String getEncodingFromOptions(final Map<?, ?> loadOptions) {
     String resultEncoding = getEncoding();
-    if (loadOptions != null && loadOptions.containsKey(XtextResource.OPTION_ENCODING)) {
+    if (loadOptions != null) {
       resultEncoding = (String) loadOptions.get(XtextResource.OPTION_ENCODING);
     }
     return resultEncoding;
```
joaodinissf, line 94: Is this equivalent?

Previously, if the map existed but didn't contain the key, this function would return `getEncoding()`. With the new version, this function could return `null` if the map exists but does not contain the key.

Are we guaranteed to get the right result because of other constraints on how the map is created/populated?

joaodinissf (changes requested):

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/XtextGMFLazyLinkingResource.java
joaodinissf: Would a `.getOrDefault()` work better?

joaodinissf (approved):

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/XtextGMFLazyLinkingResource.java
rubenporras: You are right that it was not the same.

I can however not get `.getOrDefault()` to compile. I think it is because the input map does not have any defined types, only `<?, ?>`. Therefore I have rewritten it without using that method.

PR #1065, opened 2025-01-09
Release 14.0.1.v20250109 0835 rel

ava-fred:


com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/layered/DefaultXtextTargetPlatform.java
```diff
@@ -1,84 +0,0 @@
...
- * Copyright (c) 2016 Avaloq Group AG and others.
- * All rights reserved. This program and the accompanying materials
- * are made available under the terms of the Eclipse Public License v1.0
- * which accompanies this distribution, and is available at
- * http://www.eclipse.org/legal/epl-v10.html
- *
- * Contributors:
- *     Avaloq Group AG - initial API and implementation
- *******************************************************************************/
-package com.avaloq.tools.ddk.xtext.builder.layered;
```
rubenporras, line 11: are you sure you want to do this refactor as part of a fix release? We will also need to refactor ASMD and other downstream projects.

I think should rather add the unload() method to the class in a different package and keep the changes small.

ava-fred: I need the refactor to get the DDK to build if I just cherry-pick the commits you mentioned. I should have realised that including it would have consequences for ASMD, and I agree that just manually adding the unload() method in the right class is the correct way forward.

PR #1117, opened 2025-04-23
Raise an exception when feature is null in DirectLinkingEObjectInputStream.readEObject

markdomeng:


com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/persistence/DirectLinkingEObjectInputStream.java
```diff
@@ -67,8 +67,8 @@ public EObject readEObject(final Resource context) throws IOException {
             eObject = (EObject) eObject.eGet(feature, false);
           }
         } else {
-          LOGGER.error(String.format("Failed to get EStructuralFeature for Eclass : %s, at readCompressedInt : %d", eObject.eClass(), next));
-
+          LOGGER.info(String.format("Failed to get EStructuralFeature for Eclass : %s, at readCompressedInt : %d", eObject.eClass(), next));
+          throw new IllegalStateException("feature is null"); //$NON-NLS-1$
```
rubenporras, line 71: what about removing the log and adding

throw new NullPointerException(String.format("EStructuralFeature can't be null for Eclass : %s, at readCompressedInt : %d", eObject.eClass(), next))

rubenporras (dismissed):

rubenporras:
@markdomeng , I am not sure, but you might need to suppress PMD or Checkstyle.

markdomeng:
> @markdomeng , I am not sure, but you might need to suppress PMD or Checkstyle.

rather than suppressing the warning I think it'll be better to just use IllegalStateException just as we do in all other places (the wraning states that it's not  good that we throw NPE's or something like that)

PR #1120, opened 2025-04-28
fix: load model when loading model associations.

andrewL-avlq:
The model associations are persisted as relations between model elements, so the model must exist when the associations are loaded from binary storage.

### DIFF
--- a/com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/persistence/ProxyModelAssociationsAdapter.java
+++ b/com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/persistence/ProxyModelAssociationsAdapter.java
@@ -21,6 +21,7 @@ import org.eclipse.emf.ecore.util.EcoreUtil;
 import org.eclipse.xtext.resource.persistence.StorageAwareResource;
 
 import com.avaloq.tools.ddk.xtext.modelinference.InferredModelAssociator;
+import com.avaloq.tools.ddk.xtext.resource.ResourceSetOptions;
 
 
 /**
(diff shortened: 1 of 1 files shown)


com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/persistence/ProxyModelAssociationsAdapter.java
```diff
@@ -69,6 +70,8 @@ private void ensureAssociationsLoaded() {
       if (resource.eAdapters().remove(this)) {
         DirectLinkingResourceStorageLoadable loadable = (DirectLinkingResourceStorageLoadable) ((DirectLinkingResourceStorageFacade) resource.getResourceStorageFacade()).getOrCreateResourceStorageLoadable(resource);
         try {
+          // the associations are mappings from AST model elements to inferred model elements (and back) so we cannot skip loading the model.
+          ResourceSetOptions.setSkipModel(resource.getResourceSet(), false);
```
rubenporras, line 74: should we log a warning?

rubenporras (approved):

com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/persistence/ProxyModelAssociationsAdapter.java
andrewL-avlq: I don't think so. Isn't this just a load that was missed when the 'skipModel' option was added?

rubenporras: but that is a programming error, isn't it?

PR #1283, opened 2026-03-07
build: enable UI tests on macOS via Tycho profile + JRE docs

joaodinissf:
Enables UI tests to run on macOS without modifying cross-platform `.launch` files.

## Background

SWT on Cocoa requires the Display to be created on the main thread. UI tests on macOS need `-XstartOnFirstThread`; without it they fail with `SWTException: Invalid thread access`. But the same flag is rejected as `Unrecognized VM option` by the JVM on Windows and Linux.

## Approach

Following the same pattern as Eclipse Platform's own UI test launches (`eclipse-platform/eclipse.platform`), which keep `.launch` files cross-platform and rely on per-developer JRE configuration for the macOS flag:

- **CI / `mvn verify`**: a new `macosx` profile in `ddk-parent/pom.xml`, activated on `family=mac`, sets the OSGi target environment (`macosx/cocoa/aarch64`) and appends `-XstartOnFirstThread` to `test.javaOptions`. No manual setup needed for CI.
- **Local Eclipse launches**: developers add `-XstartOnFirstThread` to their JRE's default VM arguments once (`Window > Preferences > Java > Installed JREs > Edit > Default VM arguments`). PDE launches with `append.args=true` (all of ours) inherit it on macOS. Documented in `AGENTS.md`.
- **`.launch` files are not modified**, so Windows and Linux developers see no `Unrecognized VM option` regression.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 *.xtendbin
 /*/xtend-gen/*
 !/*/xtend-gen/.gitignore
+.tycho-consumer-pom.xml
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -68,6 +68,21 @@ export WORKSPACE=$(pwd)
 - **Aggregator module**: `com.avaloq.tools.ddk.xtext.test`
 - **UI tests**: Require virtual display (xvfb on Linux) or `-XstartOnFirstThread` (macOS)
 
+### macOS developer setup
+
+SWT on Cocoa requires the Display to be created on the main thread. UI tests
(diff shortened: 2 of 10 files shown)


joaodinissf:
> **Note to self:** Test the launch configs on Windows before merging — `-XstartOnFirstThread` is macOS-specific and the JVM may reject it on other platforms. The POM profile handles this correctly for CI, but the `.launch` files hardcode the flag unconditionally.

Confirmed this doesn't work, pivoting the PR to something compatible instead.

rubenporras (approved):

PR #1299, opened 2026-04-01
fix: suppress SLF4J startup warning in Tycho test runtime

joaodinissf:
## Summary

- Remove unused `org.slf4j` imports from three test manifests
- Suppress the `SLF4J(W): No SLF4J providers were found` warning during Tycho Surefire test execution via `-Dslf4j.internal.verbosity=ERROR`

The warning is cosmetic — SLF4J defaults to NOP regardless. ServiceLoader-based provider discovery doesn't work across OSGi classloaders, which is a [known limitation](https://github.com/qos-ch/slf4j/issues/427) shared by Eclipse Platform.

## References

- [SLF4J FAQ — `slf4j.internal.verbosity`](https://www.slf4j.org/faq.html)
- [qos-ch/slf4j#427](https://github.com/qos-ch/slf4j/issues/427)
- [eclipse-platform#588](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/588)


joaodinissf:
~~This PR is stacked on #1280 (which adds `slf4j.nop` to the target platform and cleans up stale `org.slf4j` imports). See #1280 for the full investigation into why the SLF4J warning occurs in OSGi and why it cannot be resolved at the bundle level.~~

No longer stacked — #1280 was closed. This PR now carries both the manifest cleanup and the warning suppression.

rubenporras (dismissed):

joaodinissf:
**Dropped `slf4j.nop` from the target platform.** The previous version added `slf4j.nop 2.0.17` to `ddk.target`, but it served no purpose: SLF4J's ServiceLoader-based provider discovery doesn't work across OSGi classloaders, so the NOP bundle was installed but never discovered. The actual fix is the `-Dslf4j.internal.verbosity=ERROR` system property, which suppresses the warning directly. No point adding a dependency that nothing can use.

rubenporras (approved):

PR #1305, opened 2026-04-06
refactor: convert log calls to parameterized logging in test utilities

joaodinissf:
Replaces string concatenation in `LOGGER.x(...)` calls with parameterized `{}` placeholders, using `Supplier` lambdas for lazy argument evaluation. Adds NOPMD suppressions where the call site uses SwtBot's SLF4J logger (not Log4j) so the rule doesn't fire there, and adds `org.apache.logging.log4j.util` to the OSGi manifests of `test.core` and `test.ui` for the `Supplier` import.

### DIFF
--- a/com.avaloq.tools.ddk.test.ui/src/com/avaloq/tools/ddk/test/ui/swtbot/SwtBotButton.java
+++ b/com.avaloq.tools.ddk.test.ui/src/com/avaloq/tools/ddk/test/ui/swtbot/SwtBotButton.java
@@ -39,7 +39,7 @@ public class SwtBotButton extends SWTBotButton {
    */
   @Override
   public SwtBotButton click() {
-    log.debug("Clicking on {}", SWTUtils.getText(widget)); //$NON-NLS-1$
+    log.debug("Clicking on {}", SWTUtils.getText(widget)); //$NON-NLS-1$ // NOPMD GuardLogStatement
     waitForEnabled();
     notify(SWT.MouseEnter);
     notify(SWT.MouseMove);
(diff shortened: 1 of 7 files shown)


com.avaloq.tools.ddk.test.ui/src/com/avaloq/tools/ddk/test/ui/swtbot/SwtBotButton.java
```diff
@@ -39,7 +39,7 @@ public SwtBotButton(final Button button, final SelfDescribing description) {
    */
   @Override
   public SwtBotButton click() {
-    log.debug("Clicking on {}", SWTUtils.getText(widget)); //$NON-NLS-1$
+    log.debug("Clicking on {}", SWTUtils.getText(widget)); //$NON-NLS-1$ // NOPMD GuardLogStatement
```
rubenporras: this and the next should be lamdas, right?

rubenporras (changes requested):

rubenporras (dismissed):

com.avaloq.tools.ddk.test.ui/src/com/avaloq/tools/ddk/test/ui/swtbot/SwtBotButton.java
joaodinissf: Actually this `.debug()` doesn't offer this signature, unfortunately. We can go with the `if` guard.

rubenporras (approved):

PR #1308, opened 2026-04-15
build: pin the patched ANTLR generator jar, drop stale ANTLR remnants

joaodinissf:
Makes the ANTLR generator dependency self-contained and removes two stale remnants of the manual-download era.

## Pin the generator jar

The MWE2 grammar workflows need the patched ANTLR 3.2 generator. `GenerateTestLanguage.mwe2.launch` already references `com.avaloq.tools.ddk.workflow/antlr-generator-3.2.0-patch.jar` on its classpath — but the jar was never committed. In its absence, Xtext's `AntlrToolFacade` silently downloads it, **without checksum verification**, from itemis hosting that has already moved once (`download.itemis.com` → `artifacts.itemis.cloud`) and whose p2 repository is dead (HTTP 404). If that URL ever disappears, no DDK grammar can be regenerated from a fresh clone.

The jar is now committed (1.4 MB, frozen upstream since 2014, BSD-licensed) with its provenance and SHA-256 recorded in `antlr-generator-3.2.0-patch.md` alongside. The bytes were verified against the URL the Xtext 2.43 sources actually use. The `workflow` project is not a Maven module, so this has no build impact; the jar matters only for IDE-based grammar regeneration.

`GenerateCheck`/`GenerateCheckCfg` launch configs use Eclipse's default classpath and keep the auto-download fallback (cached as a gitignored hidden file); the note documents this. The jar becomes deletable if generation ever moves to `xtext-maven-plugin` + `org.eclipse.xtext:xtext-antlr-generator` from Maven Central.

## Remove stale remnants

- **`ddk-parent/readme.txt`** — "Download antlr-generator-3.2.0.jar into this folder": names the unpatched jar and a location that no resolution route (classpath, working directory, download cache) ever reads. Nothing references it.
- **`ddk-target/ddk-antlr.target`** — orphaned target definition whose primary location is the dead itemis p2 repository, pinning 2022-era Eclipse/SWTBot/Orbit URLs that contradict the maintained `ddk.target`. Zero references anywhere in the repo; not wired into the Maven/Tycho build.


rubenporras (changes requested):
you need to add the .jar to the .classpath file, otherwise it is not usable from the workspace.

rubenporras (approved):

PR #1326, opened 2026-04-30
chore: drop JUnit 4 from target and bundle manifests

joaodinissf:
No DDK source still depends on JUnit 4. This PR drops `org.junit` and `junit-vintage-engine` from the `ddk.target` IUs and from `Require-Bundle` across 7 bundle manifests, removes stale `Export-Package` entries for already-deleted packages, and switches `swtbot.eclipse.test.junit.feature.group` to `swtbot.junit5.feature.group`. `mvn clean verify` locally: 357 tests, 0 failures, 0 errors.

Two non-mechanical bits: `xtext.test`'s manifest gains an explicit `Require-Bundle` on `junit-platform-suite-engine` — the runtime executor for `@Suite`/`@SelectClasses`, previously resolved transitively by luck, and the absence of which is what made #1292's silent-zero possible. And `com.avaloq.tools.ddk.test.core.MultipleTestProblems` (the JUnit 4 implementation using `MultipleFailureException`) is deleted; its sole user `CompoundStep` switches to the existing `jupiter/MultipleTestProblems` which uses `org.opentest4j.MultipleFailuresError`.


com.avaloq.tools.ddk.xtext.test.core/META-INF/MANIFEST.MF
```diff
@@ -28,17 +27,8 @@ Require-Bundle: com.avaloq.tools.ddk.xtext,
  junit-jupiter-api
 Import-Package: org.apache.logging.log4j
 Export-Package: com.avaloq.tools.ddk.xtext.test,
- com.avaloq.tools.ddk.xtext.test.contentassist,
```
rubenporras: we should not remove this exports.

rubenporras (changes requested):
see comment about the export removal

com.avaloq.tools.ddk.xtext.test.core/META-INF/MANIFEST.MF
joaodinissf: The packages those entries point to were deleted in d920b3bad.
They had become dead code in this repo and all downstream repos following the JUnit 5 migration — platform setup in 38918bcad, per-module migrations in be881df4c, bd5dd6d11, c0dd0ad4f, 1828d2b61 and cadb87b21, final assertion cleanup in 425c26f1f.
There was one leftover trivial fix that was overlooked, which I will prepare.

KrisLimbo: This? https://github.com/dsldevkit/dsl-devkit/pull/1342/

rubenporras (approved):

PR #1333, opened 2026-05-06
ci: consolidate pmd, checkstyle, spotbugs into one static-analysis job

joaodinissf:
Replaces the separate `pmd` and `checkstyle` jobs with a single `static-analysis` job:

1. **Compile + generate reports** (`mvn compile pmd:pmd pmd:cpd checkstyle:checkstyle`) — no `*:check` goals, so no Maven cascade-skip; every module's XML is produced.
2. **Annotate + count (Python)** — parses every report, emits one inline PR annotation per violation, exits 1 if any. Fail-fast trigger that bypasses SpotBugs (slow) when there are PMD/Checkstyle issues.
3. **Maven `pmd:check pmd:cpd-check checkstyle:check` safety-net** — official-tool validation; redundant in the success case.
4. **Generate SpotBugs report** — only runs if step 3 passed.
5. **Annotate + count SpotBugs (Python)** — same pattern.
6. **Maven `spotbugs:check` safety-net.**

Drops the now-redundant pmd/checkstyle/spotbugs goals from `maven-verify` (now just `mvn clean verify`). `line-endings` stays separate. Uses `-T 2C` (8 worker threads on a 4 vCPU runner; benchmark on this branch showed ~25% wall-clock saving over sequential).

## Why every analysis step prepends `compile`

`pmd:check`, `pmd:pmd`, `spotbugs:spotbugs`, and `spotbugs:check` all need an aux-classpath populated by Tycho's compile phase. That state is in-process — even though `target/classes/` exists from a previous step, a fresh `mvn` invocation doesn't pick it up. Running the analysis goal in the same invocation as `compile` makes type-resolving rules (like PMD's `InvalidLogMessageFormat` against the SLF4J trailing-`Throwable` idiom) work correctly. Tycho's incremental compile is near no-op on the second invocation since sources haven't changed between steps.

## Cascade-skip when checks fail

With `--fail-at-end`, when `pmd:check` fails on a module, downstream modules are SKIPPED and their violations don't appear in the log either. The Python step (which parses the XML written in step 1's no-fail report run) covers everything; the Maven safety-net only sees modules that ran before any cascade-skip. In practice both layers together surface most violations on the first iteration; edge cases on cascade-skipped modules become visible after the first failing module is fixed.

### DIFF
--- a/.github/scripts/annotate-violations.py
+++ b/.github/scripts/annotate-violations.py
@@ -1,0 +1,100 @@
+#!/usr/bin/env python3
+"""Annotate PMD, Checkstyle, and SpotBugs violations as GitHub workflow commands.
+
+Walks `target/` reports under GITHUB_WORKSPACE, emits one `::warning` or `::error`
+per violation (rendered inline on the PR's Files-changed view), and exits 1 if any
+violation was found so the workflow step fails fast.
+"""
+import argparse
+import os
+import sys
+from pathlib import Path
+from xml.etree import ElementTree as ET
+
+
+def make_emit(root: Path):
+    count = 0
+
+    def emit(level: str, file, line, title: str, msg: str | None) -> None:
+        nonlocal count
+        count += 1
+        text = (msg or '').strip().replace('\n', ' ')[:1000]
+        try:
+            rel = Path(file).relative_to(root)
+        except ValueError:
+            rel = file
+        print(f"::{level} file={rel},line={line},title={title}::{text}")
+
+    return emit, lambda: count
+
+
+def parse_pmd(root: Path, emit) -> None:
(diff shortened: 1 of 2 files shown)


rubenporras (changes requested):
is it possible to not repeat the python script?

joaodinissf:
> is it possible to not repeat the python script?

I will refactor. Still want to try a few more variations before merging, I have a few ideas to optimize this further.

joaodinissf:
Superseded by #1396 (SARIF + early-fail static analysis; drops the Python annotator). The compile-in-same-invocation / cascade-skip rationale from here is carried into #1396's design doc.

*Left by Claude at João's request.*

PR #1334, opened 2026-05-06
chore: exclude PMD InvalidLogMessageFormat (SLF4J trailing-Throwable convention)

joaodinissf:
PMD's `InvalidLogMessageFormat` rule flags the SLF4J idiom `LOGGER.x("...{}...", arg, throwable)` as an argument-count mismatch (1 placeholder vs 2 args). Per [SLF4J's FAQ](https://www.slf4j.org/faq.html#paramException) this is the documented and intended way to log a parameterized message together with an exception — SLF4J trims the trailing `Throwable`, uses the remaining args for placeholders, and attaches the `Throwable` to the log event for stack-trace rendering. Excluding the rule in `ddk-configuration/pmd/ruleset.xml` removes the per-site `// NOPMD` noise that would otherwise be needed.

### DIFF
--- a/ddk-configuration/pmd/ruleset.xml
+++ b/ddk-configuration/pmd/ruleset.xml
@@ -30,6 +30,7 @@
     <exclude name="DoNotTerminateVM"/>
     <exclude name="JumbledIncrementer"/><!--Checkstyle-->
     <exclude name="EmptyCatchBlock"/><!--Checkstyle-->
+    <exclude name="InvalidLogMessageFormat"/><!--Conflicts with the SLF4J trailing-Throwable idiom `LOGGER.x("...{}...", arg, throwable)`, which SLF4J documents as the intended way to log a parameterized message with an exception. See https://www.slf4j.org/faq.html#paramException -->
     <exclude name="NullAssignment"/>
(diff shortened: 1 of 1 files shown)


rubenporras (changes requested):
This does not make much sense, according to the documentation PMD LOGGER.x("...{}...", arg, throwable) is correct: https://pmd.github.io/pmd/pmd_rules_java_errorprone.html

Is there a particular code where PMD is failing?

joaodinissf:
_(Comment left by Claude at João's request.)_

You were right — the PMD docs are accurate. Closing this in favour of the real fix in #1333.

The root cause was that our CI invoked `pmd:check` as a separate `mvn` command from `compile`, so PMD's aux-classpath was empty and `InvalidLogMessageFormat` couldn't resolve `LOGGER`'s type → false positives on the SLF4J trailing-`Throwable` idiom. Confirmed by running `mvn compile pmd:check` (single invocation) on master: 0 violations, exactly as the docs describe.

Fix in #1333: every analysis step now runs `compile` in the same `mvn` invocation as the analyzer (cheap with Tycho's incremental compile). No rule exclude needed.

PR #1353, opened 2026-05-19
chore: set line.separator=\n in source-bearing bundles (step 2/3)

### LINKED ISSUE #1345: Extend LF line-ending normalization to remaining runtime generators
## Context

#1331 added `LfNormalizingFileSystemAccess` and wired it into `CheckGenerator` and `CheckCfgGenerator` so generated output is LF regardless of host OS. The same risk pattern exists in three other runtime `IGenerator`/`IGenerator2` classes that emit Java sources to committed `src-gen/` directories — they were not addressed in #1331.

## Generators with the same risk

| Generator | Bundle | Entry point |
|---|---|---|
| `ScopeGenerator` | `com.avaloq.tools.ddk.xtext.scope` | `doGenerate(Resource, IFileSystemAccess2, IGeneratorContext)` |
| `FormatGenerator` | `com.avaloq.tools.ddk.xtext.format` | `doGenerate(Resource, IFileSystemAccess)` |
| `ExportGenerator` | `com.avaloq.tools.ddk.xtext.export` | `doGenerate(Resource, IFileSystemAccess2, IGeneratorContext)` (8 `generateFile` call sites) |

Plus an FSA-bypass writer that needs separate handling:

- `KeywordAnalysisHelper` uses raw `PrintWriter` (`.println()` → platform line separator) at lines 183, 444, 448 — writes report and documentation files to `srcGenPath`. The decorator-wrapper approach doesn't apply here; fix is either explicit `\n` or a `BufferedWriter` with explicit LF.

## Why this matters

`.gitattributes` (`* text=auto eol=lf`) normalizes on commit, so the committed state and the line-endings CI check stay green. The value is preventing **working-tree noise** for Windows developers: after regeneration, every file shows as modified in `git status`/`git diff` due to EOL drift relative to the index.

## Suggested approach

1. Hoist `LfNormalizingFileSystemAccess` from `com.avaloq.tools.ddk.check.core` to a shared bundle (e.g. `com.avaloq.tools.ddk.xtext`) so the three generators above can depend on it without pulling in `check.core`.
2. Wrap the FSA at the entry of each `doGenerate(...)` in the same pattern as #1331.
3. Decide whether `KeywordAnalysisHelper` is in scope — if yes, fix via explicit `\n` rather than the wrapper.

## Out of scope

- MWE2 `*Fragment2` classes (build-time workflow only, route through Xtext's `IXtextGeneratorFileSystemAccess` — separate path).
- Per-test generators (`TestLanguageGenerator`) — their output isn't committed.

_Left by Claude at João's request._

joaodinissf:
## What

Step 2 of the LF-alignment stack. Fixes **Eclipse-side** emission (the IDE save path).

Adds `.settings/org.eclipse.core.runtime.prefs` to 59 source-bearing bundles (every bundle with a `src/` directory that didn't already have one). Each new file is two lines:

```properties
eclipse.preferences.version=1
line.separator=\n
```

## Why

`EclipseResourceFileSystemAccess2` (the FSA Xtext uses when generating from inside Eclipse) honours the project's `line.separator` preference. Without it, the **workspace default** applies — on Windows that's CRLF — producing committed src-gen that fights `.gitattributes` and the project's LF convention.

This is the second layer of the same fix:

- **Step 1** (#1352) fixed the Maven-side (MWE2) path
- **Step 2** (this PR) fixes the Eclipse-side path
- Step 3 will then be able to remove `LfNormalizingFileSystemAccess` (PR #1331) and close #1345

Only one bundle (`ddk-configuration`) had this preference before this PR, and step 1 fixed its previously-CRLF value. This PR propagates it to the remaining 59.

## Excluded bundles (intentionally)

- `ddk-target`, `ddk-repository`, `*.feature` bundles — no `src/` directory, no code emission.

## Stack

Stacked on `chore/emit-lf-step-1` (#1352). When step 1 merges, GitHub auto-retargets this PR's base to `master`.

## Verification

- `mvn clean verify -T 3C -f ./ddk-parent/pom.xml` is expected to pass — Maven ignores Eclipse-only `.settings/` content.
- After merge of steps 1+2, a Windows reviewer regenerating any `.check` / `.checkcfg` / `.scope` / `.format` / `.export` file in Eclipse should produce LF output natively (without the `LfNormalizingFileSystemAccess` decorator). This is the precondition for step 3.

### DIFF
--- a/com.avaloq.tools.ddk.check.core.test/.settings/org.eclipse.core.runtime.prefs
+++ b/com.avaloq.tools.ddk.check.core.test/.settings/org.eclipse.core.runtime.prefs
@@ -1,0 +1,2 @@
+eclipse.preferences.version=1
+line.separator=\n
--- a/com.avaloq.tools.ddk.check.core/.settings/org.eclipse.core.runtime.prefs
+++ b/com.avaloq.tools.ddk.check.core/.settings/org.eclipse.core.runtime.prefs
@@ -1,0 +1,2 @@
+eclipse.preferences.version=1
+line.separator=\n
--- a/com.avaloq.tools.ddk.check.ide/.settings/org.eclipse.core.runtime.prefs
+++ b/com.avaloq.tools.ddk.check.ide/.settings/org.eclipse.core.runtime.prefs
@@ -1,0 +1,2 @@
+eclipse.preferences.version=1
+line.separator=\n
--- a/com.avaloq.tools.ddk.check.lib/.settings/org.eclipse.core.runtime.prefs
+++ b/com.avaloq.tools.ddk.check.lib/.settings/org.eclipse.core.runtime.prefs
@@ -1,0 +1,2 @@
+eclipse.preferences.version=1
+line.separator=\n
--- a/com.avaloq.tools.ddk.check.runtime.core.test/.settings/org.eclipse.core.runtime.prefs
+++ b/com.avaloq.tools.ddk.check.runtime.core.test/.settings/org.eclipse.core.runtime.prefs
@@ -1,0 +1,2 @@
+eclipse.preferences.version=1
+line.separator=\n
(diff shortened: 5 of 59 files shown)


andrewL-avlq:
The additional preferences files should not be needed as they already exist as links in the workspace to the preferences in ddk-configuration.

PR #1360, opened 2026-05-22
feat: publish p2 update site to GitHub Pages via gh-pages branch

rubenporras:
- Create a snapshot workflow that build and deploys on every push to master or v[0-9]* branches into p2/snapshots/<sha8>/; keep last 20, clean up older ones
- Create a release workflow to promote existing snapshots to releases; The workflow creates a tag and  copy the corresponding snapshot for that commit into p2/releases/<version>/ on gh-pages
- Add composite p2 repositories (compositeContent.xml, compositeArtifacts.xml, p2.index) for p2/releases/latest/ and p2/snapshots/latest/ so Eclipse can resolve "latest" without a fixed URL
- Add generated index.html listing all snapshot and release links
- Document p2 update site URLs in README.md

### DIFF
--- a/.github/scripts/cleanup-p2-snapshots.sh
+++ b/.github/scripts/cleanup-p2-snapshots.sh
@@ -1,0 +1,43 @@
+#!/usr/bin/env bash
+# Removes old p2 snapshots from gh-pages, keeping the last 20.
+# Ordering uses the .created timestamp written by publish-p2-ghpages.sh;
+# dirs without that file (pre-dating this change) are treated as oldest.
+# Required env vars: GITHUB_TOKEN
+
+set -euo pipefail
+
+KEEP=20
+
+cd gh-pages
+git config user.email "github-actions[bot]@users.noreply.github.com"
+git config user.name "github-actions[bot]"
+
+if [ -d p2/snapshots ]; then
+  to_delete=$(
+    for d in $(ls -1 p2/snapshots | grep -v latest); do
+      ts=0
+      [ -f "p2/snapshots/$d/.created" ] && ts=$(cat "p2/snapshots/$d/.created")
+      printf '%s %s\n' "$ts" "$d"
+    done | sort -n | head -n -"$KEEP" | awk '{print $2}'
+  )
+  if [ -n "$to_delete" ]; then
+    echo "$to_delete" | xargs -I{} rm -rf "p2/snapshots/{}"
+  fi
+fi
+
+git add -A
+if git diff --cached --quiet; then
+  echo "No snapshot cleanup needed"
+  exit 0
+fi
+git commit -m "Clean up old p2 snapshots"
+for attempt in 1 2 3; do
+  if git push origin HEAD:gh-pages; then
+    exit 0
+  fi
+  echo "Push attempt $attempt failed; fetching and rebasing"
+  git fetch origin gh-pages
+  git rebase origin/gh-pages
+done
+echo "Failed to push gh-pages after 3 attempts"
+exit 1
--- a/.github/scripts/publish-p2-ghpages.sh
+++ b/.github/scripts/publish-p2-ghpages.sh
@@ -1,0 +1,126 @@
+#!/usr/bin/env bash
+# Publishes the p2 update site to the gh-pages branch.
+#
+# Snapshot mode (RELEASE_VERSION unset):
+#   Copies ddk-repository/target/repository/ → p2/snapshots/<sha8>/
+#   Updates p2/snapshots/latest/ composite repository
+#
+# Release mode (RELEASE_VERSION set):
+#   Copies p2/snapshots/<SNAPSHOT_SHA>/ → p2/releases/<RELEASE_VERSION>/
+#   Updates p2/releases/latest/ composite repository
+#
+# Required env vars: GITHUB_TOKEN, GITHUB_SHA, GITHUB_REPOSITORY
+# Release-only vars: RELEASE_VERSION, SNAPSHOT_SHA
+
+set -euo pipefail
+
+write_composite() {
+  local dir="$1"
+  local name="$2"
+  local child="$3"
+  rm -rf "$dir"
+  mkdir -p "$dir"
+  cat > "$dir/p2.index" << 'EOF'
+version=1
+metadata.repository.factory.order=compositeContent.xml,\!
+artifact.repository.factory.order=compositeArtifacts.xml,\!
+EOF
+  cat > "$dir/compositeContent.xml" << EOF
+<?xml version='1.0' encoding='UTF-8'?>
+<?compositeMetadataRepository version='1.0.0'?>
+<repository name='$name' type='org.eclipse.equinox.internal.p2.metadata.repository.CompositeMetadataRepository' version='1.0.0'>
+  <properties size='1'><property name='p2.timestamp' value='$TIMESTAMP'/></properties>
+  <children size='1'><child location='$child'/></children>
+</repository>
+EOF
+  cat > "$dir/compositeArtifacts.xml" << EOF
+<?xml version='1.0' encoding='UTF-8'?>
+<?compositeArtifactRepository version='1.0.0'?>
+<repository name='$name' type='org.eclipse.equinox.internal.p2.artifact.repository.CompositeArtifactRepository' version='1.0.0'>
+  <properties size='1'><property name='p2.timestamp' value='$TIMESTAMP'/></properties>
+  <children size='1'><child location='$child'/></children>
+</repository>
+EOF
+}
+
+cd gh-pages
+git config user.email "github-actions[bot]@users.noreply.github.com"
+git config user.name "github-actions[bot]"
+
+touch .nojekyll
+
+if [ -n "${RELEASE_VERSION:-}" ]; then
+  TARGET="p2/releases/$RELEASE_VERSION"
+  REF_DESC="release $RELEASE_VERSION"
+  rm -rf "$TARGET"
+  mkdir -p "$TARGET"
+  cp -r "p2/snapshots/$SNAPSHOT_SHA/." "$TARGET/"
+else
+  TARGET="p2/snapshots/$SNAPSHOT_SHA"
+  REF_DESC="snapshot $SNAPSHOT_SHA"
+  rm -rf "$TARGET"
+  mkdir -p "$TARGET"
+  cp -r ../ddk-repository/target/repository/. "$TARGET/"
+  date +%s > "$TARGET/.created"
+fi
+
+TIMESTAMP="$(date +%s)000"
+if [ -n "${RELEASE_VERSION:-}" ]; then
+  HIGHEST=$(ls -1 p2/releases | grep -v latest | sort -Vr | head -1)
+  write_composite p2/releases/latest "DDK p2 Latest Release" "../$HIGHEST/"
+else
+  write_composite p2/snapshots/latest "DDK p2 Latest Snapshot" "../$SNAPSHOT_SHA/"
+fi
+
+# Generate index.html
+{
+  echo '<!DOCTYPE html>'
+  echo '<html><body>'
+  echo '<h1>DSL DevKit p2 Repository</h1>'
+  echo '<h2>Snapshots</h2>'
+  echo '<ul>'
+  if [ -d p2/snapshots/latest ]; then
+    echo '  <li><a href="p2/snapshots/latest/">Latest Snapshot</a></li>'
+  fi
+  if [ -d p2/snapshots ]; then
+    while IFS= read -r s; do
+      echo "  <li><a href=\"p2/snapshots/$s/\">$s</a></li>"
+    done < <(
+      for d in $(ls -1 p2/snapshots | grep -v latest); do
+        ts=0
+        [ -f "p2/snapshots/$d/.created" ] && ts=$(cat "p2/snapshots/$d/.created")
+        printf '%s %s\n' "$ts" "$d"
+      done | sort -rn | head -20 | awk '{print $2}'
+    )
+  fi
+  echo '</ul>'
+  if [ -d p2/releases ] && [ -n "$(ls -A p2/releases 2>/dev/null)" ]; then
+    echo '<h2>Releases</h2>'
+    echo '<ul>'
+    if [ -d p2/releases/latest ]; then
+      echo '  <li><a href="p2/releases/latest/">Latest Release</a></li>'
+    fi
+    for v in $(ls -1 p2/releases | grep -v latest | sort -Vr); do
+      echo "  <li><a href=\"p2/releases/$v/\">$v</a></li>"
+    done
+    echo '</ul>'
+  fi
+  echo '</body></html>'
+} > index.html
+
+git add -A
+if git diff --cached --quiet; then
+  echo "No changes to publish"
+  exit 0
+fi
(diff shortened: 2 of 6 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,158 @@
...
+            minor) MINOR=$((MINOR+1)); PATCH=0 ;;
+            patch) PATCH=$((PATCH+1)) ;;
+          esac
+          NEW_TAG="v${MAJOR}.${MINOR}.${PATCH}"
+          echo "tag=$NEW_TAG" >> "$GITHUB_OUTPUT"
+          echo "Previous version $LATEST → next version $NEW_TAG"
+
+      - name: Set snapshot SHA
+        id: snapshot
+        run: echo "sha=${GITHUB_SHA::8}" >> "$GITHUB_OUTPUT"
```
joaodinissf: **`GITHUB_SHA` is not the SHA of the checked-out branch.**

On `workflow_dispatch`, `GITHUB_SHA` is the SHA of the **ref the workflow was dispatched on** (usually `master`), not `inputs.branch`. If a user dispatches from the master selector but picks `inputs.branch: release/1.0.x`, the previous step checked out `release/1.0.x` but `snapshot_sha` here is `master`'s HEAD. The verify-snapshot step then looks up the wrong snapshot — best case it 404s with a confusing error, worst case the master snapshot exists and we promote unrelated bits as `release/1.0.x`.

Derive from the checkout instead:
```yaml
run: echo "sha=$(git rev-parse --short=8 HEAD)" >> "$GITHUB_OUTPUT"
```

_left by Claude at João's request_

joaodinissf (commented):
Review of the publish-p2 pipeline. Overall the structure is clean — concurrency group, gh-pages bootstrap, push-with-rebase loop, pinned actions all look good. A few issues worth addressing before merge, flagged inline below.

_left by Claude at João's request_

ddk-parent/pom.xml
```diff
@@ -215,7 +222,17 @@
...
+        <dependencies>
+          <dependency>
+            <groupId>org.eclipse.tycho</groupId>
+            <artifactId>tycho-buildtimestamp-jgit</artifactId>
+            <version>${tycho.version}</version>
+          </dependency>
+        </dependencies>
         <configuration>
+          <timestampProvider>jgit</timestampProvider>
+          <jgit.ignore>pom.xml</jgit.ignore>
```
joaodinissf: **Correction to my earlier comment on this line — original claim was wrong.** I asserted `jgit.ignore` was a comma-separated list of literal paths matching only the root `pom.xml`. Verified against [tycho-extras PathFilter.java](https://github.com/eclipse-tycho/tycho/blob/main/tycho-buildtimestamp-jgit/src/main/java/org/eclipse/tycho/extras/buildtimestamp/jgit/PathFilter.java): the value is **newline-split** and each line is fed to JGit's `FastIgnoreRule`, i.e. real gitignore semantics. A bare `pom.xml` matches *every* `pom.xml` at any depth — exactly what one would want for an across-the-tree exemption.

With that out of the way, the substantive question is whether this exemption is correctly *scoped*. The default exempted set (hardcoded in `PathFilter`) is already `.tycho-consumer-pom.xml`, `.polyglot.*`, `pom.tycho` — Tycho-generated files. Adding `pom.xml` is a deliberate choice that says "expect human-edited `pom.xml` to be dirty at build time."

In the current PR, neither `snapshot.yml` nor `release.yml` modifies any pom mid-build, so the practical effect today is zero. The presumable forward-looking intent is the "automatic Bump PR" workflow alluded to in `release.yml`'s trailing comment, which would run `tycho-versions:set-version`. If that's the case, the exemption is **incomplete** — `tycho-versions:set-version` also rewrites:

- `META-INF/MANIFEST.MF` (`Bundle-Version`)
- `feature.xml`
- `category.xml`
- `*.product`

so the dirty-tree check would still fire on those.

Suggested directions, in rough order of cleanliness:

1. **Commit the bump, then build.** Have the bump PR's CI run `tycho-versions:set-version`, commit on the PR branch, then build/test from a clean tree. No `jgit.ignore` extension needed at all; gives a reviewable diff. This is what most Tycho projects do.
2. **If in-build bumps are genuinely required**, expand `jgit.ignore` to cover everything `tycho-versions` writes (newline-separated):
   ```xml
   <jgit.ignore>
     pom.xml
     MANIFEST.MF
     feature.xml
     category.xml
     *.product
   </jgit.ignore>
   ```
3. **If `pom.xml` is here for some reason I haven't seen**, an XML comment explaining the case would help future readers.

Not a blocker — just worth confirming the intended scope.

_left by Claude at João's request_

ddk-parent/pom.xml
```diff
@@ -56,6 +56,13 @@
     <pmd.version>7.24.0</pmd.version>
     <tycho.version>5.0.2</tycho.version>
     <xtend.version>2.42.0</xtend.version>
+
+    <!-- jgit qualifier: error in CI, warning for local dirty trees -->
```
joaodinissf, line 60: **Comment doesn't match behavior — and the lsp4e-pattern fix is concrete.**

lsp4e (whose pom is clearly the template here) implements exactly what this comment promises with an `!env.CI`-activated profile. This PR has the comment and the `error` property but is missing the profile, so any local `mvn verify` with uncommitted changes will fail dirty.

Suggested fix for this line:

```suggestion
    <!-- jgit qualifier dirty-tree handling: error in CI; warning locally via the !env.CI profile below -->
```

And add this profile in the existing `<profiles>` section (around line 423):

```xml
<profile>
  <id>local-build</id>
  <activation>
    <property>
      <name>!env.CI</name>
    </property>
  </activation>
  <properties>
    <jgit.dirtyWorkingTree>warning</jgit.dirtyWorkingTree>
  </properties>
</profile>
```

GitHub Actions sets `CI=true`, so the profile only activates outside CI. Matches lsp4e verbatim.

_left by Claude at João's request_

ddk-parent/pom.xml
```diff
@@ -56,6 +56,13 @@
...
+    <jgit.dirtyWorkingTree>error</jgit.dirtyWorkingTree>
+
+    <!-- Baseline version validation (skip until first release populates p2/releases/latest/) -->
+    <baseline.repo.url>https://dsldevkit.github.io/dsl-devkit/p2/releases/latest/</baseline.repo.url>
+    <baseline.skip>true</baseline.skip>
```
joaodinissf, line 65: **Follow-up tracking?** `baseline.skip=true` is correct for now (no `p2/releases/latest/` exists until the first release publishes), but it needs to be flipped to `false` once the first release lands — otherwise baseline version validation is silently disabled forever. Worth filing a follow-up issue, or wiring the property to a profile that auto-disables for the very first release only.

_left by Claude at João's request_

.github/workflows/release.yml
```diff
@@ -0,0 +1,158 @@
...
+            echo "| **Tag** | [$TAG]($REPO/releases/tag/$TAG) |"
+            echo "| **Branch** | \`${{ inputs.branch }}\` |"
+            echo "| **Bump** | \`${{ inputs.version_bump }}\` |"
+          } >> "$GITHUB_STEP_SUMMARY"
+
+# Release workflow:
+#   Snapshots:  every push to master or v[0-9]* branches → see snapshot.yml
+#   Release:    workflow_dispatch → create tag, promote snapshot to p2/releases/<version>/
+#   Bump:       automatic PR after each release (0-4 minor cycle, then major)
+#   Patches:    run from release/X.Y.x branch → bump PR targets release branch
```
joaodinissf: **Trailing comment describes workflows that don't exist in this PR.**

```
#   Bump:       automatic PR after each release (0-4 minor cycle, then major)
#   Patches:    run from release/X.Y.x branch → bump PR targets release branch
```

There's no bump-PR workflow, no `release/X.Y.x` patch flow. Either drop these lines or label them as "future" so reviewers/users don't go looking for them.

Also: file has no trailing newline.

_left by Claude at João's request_

.github/scripts/cleanup-p2-snapshots.sh
```diff
@@ -0,0 +1,30 @@
...
+# Required env vars: GITHUB_TOKEN
+
+set -euo pipefail
+
+cd gh-pages
+git config user.email "github-actions[bot]@users.noreply.github.com"
+git config user.name "github-actions[bot]"
+
+if [ -d p2/snapshots ]; then
+  (cd p2/snapshots && ls -1t | grep -v latest | tail -n +21 | xargs -r -I{} rm -rf "{}")
```
joaodinissf: **Retention is not actually "keep the newest 20" — it's "keep the newest 1 and 19 arbitrary others".**

`actions/checkout` does not preserve git commit times as filesystem mtimes; every file/dir on a fresh checkout gets mtime ≈ checkout time. So when this runs on a workflow with >20 existing snapshots:

- The just-`cp -r`'d new snapshot has a fresh mtime (newest).
- All previously-existing snapshot dirs share the same mtime (the checkout time), within milliseconds.
- `ls -1t` ties are broken by filesystem ordering, which is undefined.

Result: on every cleanup run, *which* old snapshots are pruned is non-deterministic, and over time the oldest snapshots are not preferentially removed.

More reliable options:
- Drive retention from git history: `git log --diff-filter=A --name-only -- 'p2/snapshots/*' | …` to find the introducing commit for each dir, sort by commit date.
- Or stamp each new snapshot dir with a timestamp file at creation, and sort by that.
- Or commit a small `p2/snapshots/index.json` listing snapshots in order, and prune from the tail.

_left by Claude at João's request_

ddk-parent/pom.xml
```diff
@@ -242,6 +259,26 @@
...
+            </goals>
+            <configuration>
+              <baselines>
+                <url>${baseline.repo.url}</url>
+              </baselines>
+              <skip>${baseline.skip}</skip>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
```
joaodinissf, line 292: **Missing baseline-replacement plugin — `compare-version-with-baselines` above will false-positive on every unchanged-module rebuild without it.**

The extras plugin defaults `<comparator>` to `bytes` (literal `FileUtils.contentEquals`). Tycho-built jars are not byte-reproducible across rebuilds out of the box: MANIFEST.MF carries `Bnd-LastModified`, `META-INF/maven/.../pom.properties` carries a build timestamp, jar entry ordering can drift, etc. So when the check finds a module at the same fully-qualified version as baseline and byte-compares, it will fail with `Baseline and reactor have the same fully qualified version, but different content`.

[lsp4e](https://github.com/eclipse-lsp4e/lsp4e/blob/main/pom.xml) addresses this by configuring `tycho-p2-plugin` with `<baselineRepositories>` pointing at the **snapshots** URL. At packaging time Tycho compares each built bundle against the baseline; where content matches semantically, the baseline jar bytes are reused wholesale. The subsequent `compare-version-with-baselines` byte comparison then trivially passes for unchanged modules.

Two URLs, two purposes:
- `<baselineRepositories>` on `tycho-p2-plugin` → snapshots (byte-identical rebuilds across CI runs)
- `<baselines>` on `tycho-p2-extras-plugin` (above) → releases (version-monotonicity check)

Suggested addition — replaces the existing `</plugin>` on this line with the same closing tag plus a new sibling plugin entry:

```suggestion
      </plugin>
      <plugin>
        <groupId>org.eclipse.tycho</groupId>
        <artifactId>tycho-p2-plugin</artifactId>
        <version>${tycho.version}</version>
        <configuration>
          <baselineRepositories>
            <repository>
              <url>https://dsldevkit.github.io/dsl-devkit/p2/snapshots/latest/</url>
            </repository>
          </baselineRepositories>
        </configuration>
      </plugin>
```

Note: this is required only if you remove `<baseline.skip>true</baseline.skip>` in the follow-up; while skip is true the comparator never runs and this is dormant. But since the plan is to unskip immediately, it should land in this PR or in the same follow-up that flips the skip.

_left by Claude at João's request_

ddk-parent/pom.xml
rubenporras: done

ddk-parent/pom.xml
rubenporras: done

joaodinissf:
Hi @rubenporras, I think the Actions issue we saw this afternoon was due to a GH outage, can you try to retrigger? Pushing an empty commit or a small change should do it.
Actually I think I can do this for you, I'll give it a try.

joaodinissf:
Sorry for the noise — closed and reopened just now to retrigger CI. The current head SHA `82914be` had no workflow runs (last build on this branch was 2026-05-22 for an older SHA, so a webhook was dropped somewhere). No code changes, no force-push, review state preserved. A new run started on reopen.

_Comment left by Claude at João's request._

.github/scripts/publish-p2-ghpages.sh
```diff
@@ -0,0 +1,127 @@
...
+touch .nojekyll
+
+if [ -n "${RELEASE_VERSION:-}" ]; then
+  TARGET="p2/releases/$RELEASE_VERSION"
+  REF_DESC="release $RELEASE_VERSION"
+  rm -rf "$TARGET"
+  mkdir -p "$TARGET"
+  cp -r "p2/snapshots/$SNAPSHOT_SHA/." "$TARGET/"
+else
+  TARGET="p2/snapshots/${GITHUB_SHA::8}"
```
joaodinissf: **Same `GITHUB_SHA` bug as the (now-fixed) `Set snapshot SHA` step in `release.yml`.**

On `workflow_dispatch` with `inputs.branch` overridden, `GITHUB_SHA` is the dispatch ref (usually `master`), not the checked-out branch's HEAD. The snapshot directory gets named after the wrong commit, and `release.yml`'s verify-snapshot step then can't find it.

Mirror the release.yml pattern — `SNAPSHOT_SHA` becomes the canonical name in both modes (it already means "8-char SHA of the snapshot being operated on" in release mode):

1. **`.github/workflows/snapshot.yml`** — after the checkout step (~L27), add:
   ```yaml
         - name: Set snapshot SHA
           id: snapshot
           run: echo "sha=$(git rev-parse --short=8 HEAD)" >> "$GITHUB_OUTPUT"
   ```
2. **`.github/workflows/snapshot.yml`** — on the `Publish p2 snapshot to gh-pages` step's `env:` (L85-86), add:
   ```yaml
             SNAPSHOT_SHA: ${{ steps.snapshot.outputs.sha }}
   ```
3. **`.github/scripts/publish-p2-ghpages.sh`** — in the snapshot-mode branch, replace `${GITHUB_SHA::8}` with `$SNAPSHOT_SHA` at L59-60, collapse `SHORT_SHA` at L72-73 to use `$SNAPSHOT_SHA` directly, and update the `Required env vars:` line at L12 (swap `GITHUB_SHA` → `SNAPSHOT_SHA`).

Optional same-root-cause cosmetic: `snapshot.yml:58` artifact name uses `${{ github.sha }}` — flip to `${{ steps.snapshot.outputs.sha }}` for consistency. The artifact isn't consumed downstream so this is naming-only.

_left by Claude at João's request_

joaodinissf (changes requested):

.github/scripts/publish-p2-ghpages.sh
rubenporras: done

rubenporras:
just rebased

joaodinissf (approved):
LGTM.

PR #1371, opened 2026-05-27
build: enable the Tycho baseline version

rubenporras:
This commit re-enables the Tycho baseline version check now that the
first v18.0.0 release is published to GitHub Pages, and adjusts the
project's version layout to support that workflow going forward.

In pom.xml, the parent POM coordinate is bumped to 18.0.1-SNAPSHOT so
the build aggregator moves ahead of the published release stream. A new
target.platform.version property is introduced and
target-platform-configuration switches from ${project.version} to
${target.platform.version}. This decouples the target platform from the
bundle versions, so future parent bumps don't force every bundle to be
re-released. 

A separate snapshot.repo.url property is added, and the
tycho-p2-plugin baselineRepositories configuration is moved into
<pluginManagement> (matching the lsp4e pattern). The temporary
<baseline.skip>true</baseline.skip> workaround is removed so
compare-version-with-baselines runs on every build.

All child plugin/feature poms are updated to reference the new
parent (18.0.1-SNAPSHOT) and now carry an explicit <version> element so
each bundle keeps its own coordinate independent of the parent. This is
required because Tycho needs the pom version to align with the MANIFEST
Bundle-Version.

Because the PR modifies every bundle's pom.xml, the jgit qualifier picks
up today's commit timestamp, which the re-enabled baseline check
correctly rejects against the already-published 17.3.0.qualifier bundles
("Only qualifier changed… expected bigger x.y.z"). To satisfy semver,
all MANIFEST.MF files, both feature.xml files
(com.avaloq.tools.ddk.feature, com.avaloq.tools.ddk.runtime.feature),
the child pom <version> elements, and category.xml are bumped from
17.3.0 to 17.3.1. ddk-parent, ddk-target, and ddk-repository remain at
18.0.1-SNAPSHOT. Exceptionally the pom.xml of 
com.avaloq.tools.ddk.workflow has been removed, as it isn't in ddk-parent's
<modules> and therefore not part of the CI/delivery.

From now on the parent POM coordinate is decoupled from the bundle
stream, the baseline check is active and protects against accidental
re-publishing of released bundles without a version bump, and all
bundles move to 17.3.1 to reflect the content change introduced by commit.


com.avaloq.tools.ddk.workflow/pom.xml
```diff
@@ -3,7 +3,7 @@
   <parent>
     <artifactId>ddk-parent</artifactId>
     <groupId>com.avaloq.tools.ddk</groupId>
-    <version>17.3.0-SNAPSHOT</version>
+    <version>17.3.1-SNAPSHOT</version>
```
joaodinissf: `17.3.1` landed on the parent ref here instead of as the module's own `<version>` (which is missing). Every other module is `(parent 18.0.1, own 17.3.1)`. Since `ddk-parent` is now `18.0.1-SNAPSHOT`, this parent won't resolve — it only escapes CI because `workflow` isn't in `ddk-parent`'s `<modules>`. Expected:
```xml
<parent>
  ...
  <version>18.0.1-SNAPSHOT</version>
</parent>
<version>17.3.1-SNAPSHOT</version>
```
Verified it's the only such deviation across the diff.

*Left by Claude at João's request.*

joaodinissf (changes requested):

com.avaloq.tools.ddk.workflow/pom.xml
rubenporras: Good catch. I have however removed the pom.xml instead of adapting it.

joaodinissf (approved):

PR #1394, opened 2026-05-30
fix: minor robustness/logging cleanups from the inconsistency sweep

joaodinissf:
Two low-risk cleanups from the repo-wide inconsistency sweep (bundled as one small PR):

- **`ScopeResourceDescriptionStrategy.getSourceText`** — guard a null `getParseResult()` before `getRootNode()` (node-model-less resources exist, e.g. direct-linking storage), returning `""` as it already does for the proxy case. Without the guard the NPE is swallowed by the enclosing `catch (RuntimeException)` and the scope model is silently not indexed at all; with it, the model is indexed with the source text contributing as empty.
- **`CustomClassAwareEcoreGenerator`** — non-OK EMF generation diagnostics were logged at `INFO`, so warnings/errors were easy to miss; log them at `WARN`.

### DIFF
--- a/com.avaloq.tools.ddk.xtext.scope/src/com/avaloq/tools/ddk/xtext/scope/resource/ScopeResourceDescriptionStrategy.java
+++ b/com.avaloq.tools.ddk.xtext.scope/src/com/avaloq/tools/ddk/xtext/scope/resource/ScopeResourceDescriptionStrategy.java
@@ -70,6 +70,11 @@ public class ScopeResourceDescriptionStrategy extends DefaultResourceDescription
       LOG.warn("Could not resolve scope model " + EcoreUtil.getURI(obj));
       return "";
     }
-    return ((XtextResource) obj.eResource()).getParseResult().getRootNode().getText();
+    final XtextResource resource = (XtextResource) obj.eResource();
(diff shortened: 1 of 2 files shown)


com.avaloq.tools.ddk.xtext.scope/src/com/avaloq/tools/ddk/xtext/scope/resource/ScopeResourceDescriptionStrategy.java
```diff
@@ -70,6 +70,10 @@ private String getSourceText(final EObject obj) {
       LOG.warn("Could not resolve scope model " + EcoreUtil.getURI(obj));
       return "";
     }
-    return ((XtextResource) obj.eResource()).getParseResult().getRootNode().getText();
+    final XtextResource resource = (XtextResource) obj.eResource();
+    if (resource.getParseResult() == null) {
+      return "";
```
rubenporras, line 76: Could we log `LOG.warn("No parse result for scope model " + EcoreUtil.getURI(obj));`

rubenporras (changes requested):

rubenporras (approved):

PR #1397, opened 2026-05-30
chore: enable CPD detection (pmd.cpd.min → 100), de-dup and baseline

### LINKED ISSUE #1339: chore: pmd.cpd.min set to 100000 effectively disables CPD
## Observation

`ddk-parent/pom.xml` sets:

```xml
<pmd.cpd.min>100000</pmd.cpd.min>
```

This is the `minimumTokens` threshold for PMD's Copy/Paste Detector (CPD). At this value, CPD is effectively disabled — `pmd:cpd-check` will pass on virtually any duplication.

## Empirical evidence

Modified `com.avaloq.tools.ddk.test.core/src/.../PostconditionViolation.java` to add two identical 8-line methods (~65 tokens each by CPD's count):

```java
public int compute1(int x) {
  int a = x + 1;
  int b = a * 2;
  int c = b - 3;
  int d = c + 4;
  int e = d * 5;
  int f = e - 6;
  return f + a + b + c + d + e;
}

public int compute2(int x) {
  // identical body to compute1
}
```

Running `pmd:cpd-check` at the project's configured threshold:

| Run | Threshold | Reactor duplications detected |
|---|---|---|
| Current `pmd.cpd.min=100000` | 100 000 tokens | **0** (synthetic 65-token duplication missed) |
| `-Dpmd.cpd.min=20` (override) | 20 tokens | **21** including the synthetic one and real ones in the codebase |

The largest pre-existing duplication detected at the lower threshold is 90 lines / 267 tokens — meaningful duplication that the project currently doesn't see.

## Suggested fix

PMD's CPD default is 100 tokens. Common production values for Java sit between 50 and 100. A reasonable starting point:

```xml
<pmd.cpd.min>100</pmd.cpd.min>
```

Then run `pmd:cpd-check` against the reactor, triage the surfaced duplications (some may be legitimate — e.g., generated code, parser tables — and warrant suppression rather than fixing), and tune from there.

## Why this matters now

- Contributors and reviewers may believe CPD enforcement is active when it's actually a no-op.
- CI redesign work (e.g. SARIF + GitHub Code Scanning experimentation) treats `pmd:cpd-check` as a gate, but it never fires today.
- 21 real duplications in the reactor are silently accumulating.

---
_Filed by Claude at João's request._

joaodinissf:
Closes #1339.

## Problem
`ddk-parent/pom.xml` set `pmd.cpd.min=100000`, which effectively **disables** PMD's copy/paste detector — `pmd:cpd-check` passes on virtually any duplication (#1339).

## What this does
- Lowers `pmd.cpd.min` to **100** (PMD's default), so CPD actually detects duplication going forward.
- At 100 the reactor surfaces **pre-existing duplications in 8 file-groups**. Each was assessed for whether it is genuinely extractable logic or incidental/deliberate similarity:
  - **2 refactored** (C, E) — genuine duplicated logic; behavior-neutral extraction.
  - **6 baselined** (A, B, D, F, G, H) with `// CPD-OFF` markers carrying a per-site reason.
- **Verified:** full local gate (`clean verify checkstyle:check pmd:pmd pmd:cpd pmd:check pmd:cpd-check spotbugs:check`) → BUILD SUCCESS, `pmd:cpd-check` clean across the full reactor at threshold 100.

## Dispositions

| Group | File(s) | Kind | Disposition | Rationale |
|---|---|---|---|---|
| **A** | `AbstractValidationTest` (test core) | test scaffolding | **baseline** | error-message assembly + assert helpers in a test base; per-test clarity beats a shared extraction that would obscure intent. |
| **B** | `ParameterListMatcherTest` (typesystem test) | test cases | **baseline** | sibling `testForceMatchByPosition*` cases share an arrange-act-assert skeleton, kept explicit for readability. |
| **C** | `Check{Validator,Quickfix}ExtensionHelper` (prod UI) | helper pair | **refactored** | identical `isExtensionUpdateRequired` guard chain lifted to `AbstractCheckExtensionHelper`, keyed on the existing `getExtensionPointId()` plus a new `protected getTargetClassName()` hook. The documentation-helper subtree (no target class) declares that hook unsupported once in its shared parent. |
| **D** | `DispatchingCheckImpl` (prod runtime) | incidental | **baseline** | the matched spans are incidental token overlap (DI field declarations + the trace/try/run/catch/finally idiom), not an extractable unit. |
| **E** | `Abstract{,Streaming}FingerprintComputer` (prod) | helper pair | **refactored** | the identical private `featureIterable` helper extracted to a shared package-private `FingerprintFeatures` utility. |
| **F** | `FormatJvmModelInferrer` (prod, migrated) | dispatch/emission | **baseline (file-scoped)** | 9 self-duplications across the per-locator emission blocks of freshly-migrated Xtend dispatch code, kept faithful to the migration; de-duplication belongs to the migration readability follow-ups, not this PR. |
| **G** | `FormatFragment2` (prod generator, migrated) | stub emission | **baseline** | the parallel Xtend- and Java-formatter stub emission runs are deliberately explicit twins. |
| **H** | `CheckCfg{ContentAssist,ConfiguredParameterValidations}Test` | test scaffolding | **baseline** | sibling test classes share `@Override` setup scaffolding, kept explicit per class. |

## Notes
- The lifted `isExtensionUpdateRequired` overrides keep their `public` visibility: the hook is deliberately public so tests can exercise it directly on injected helpers; #1460 makes the whole hierarchy uniformly public.
- CPD has no SARIF renderer, so it gates (the build goes red on a duplication) but doesn't annotate inline; a `cpd.xml`→SARIF transform is a possible follow-up.
- On the current workflow CPD gates via `maven-verify`'s `pmd:cpd-check`; once #1396's lint lane lands, its CPD gate becomes meaningful too (its "threshold currently very high" note can then be dropped).

**Merge chain:** ~~#1399~~ (merged) → #1396 → #1400 → #1456 → #1457 → ***#1397 (this PR)***

### DIFF
--- a/com.avaloq.tools.ddk.xtext.format.generator/src/com/avaloq/tools/ddk/xtext/format/generator/FormatFragment2.java
+++ b/com.avaloq.tools.ddk.xtext.format.generator/src/com/avaloq/tools/ddk/xtext/format/generator/FormatFragment2.java
@@ -164,6 +164,7 @@ public class FormatFragment2 extends AbstractStubGeneratingFragment {
       protected void appendTo(final TargetStringConcatenation builder) {
         builder.append("import com.avaloq.tools.ddk.xtext.formatting.ExtendedLineEntry");
         builder.newLine();
+        // CPD-OFF — parallel Xtend/Java formatter-stub emission, kept explicit (#1339)
         builder.append("import java.util.List");
         builder.newLine();
         builder.newLine();
@@ -184,6 +185,7 @@ public class FormatFragment2 extends AbstractStubGeneratingFragment {
         builder.append(" */");
         builder.newLine();
         builder.append("class ");
+        // CPD-ON
         builder.append(getFormatterStub(getGrammar()).getSimpleName());
         builder.append(" extends ");
         builder.append(FormatGeneratorUtil.getFormatterName(getGrammar(), "Abstract"));
@@ -251,6 +253,7 @@ public class FormatFragment2 extends AbstractStubGeneratingFragment {
         builder.append("import java.util.List;");
         builder.newLine();
         builder.newLine();
+        // CPD-OFF — parallel Xtend/Java formatter-stub emission, kept explicit (#1339)
         builder.append("import org.eclipse.xtext.TerminalRule;");
         builder.newLine();
         builder.newLine();
@@ -271,6 +274,7 @@ public class FormatFragment2 extends AbstractStubGeneratingFragment {
         builder.append(" */");
         builder.newLine();
         builder.append("public class ");
+        // CPD-ON
         builder.append(getFormatterStub(getGrammar()).getSimpleName());
         builder.append(" extends ");
         builder.append(FormatGeneratorUtil.getFormatterName(getGrammar(), "Abstract"));
--- a/com.avaloq.tools.ddk.check.runtime.core/src/com/avaloq/tools/ddk/check/runtime/issue/DispatchingCheckImpl.java
+++ b/com.avaloq.tools.ddk.check.runtime.core/src/com/avaloq/tools/ddk/check/runtime/issue/DispatchingCheckImpl.java
@@ -35,6 +35,7 @@ import com.google.inject.Inject;
 @SuppressWarnings({"checkstyle:AbstractClassName"})
 public abstract class DispatchingCheckImpl extends AbstractCheckImpl {
 
+  // CPD-OFF — incidental token overlap (DI field + trace/try idiom), not an extractable unit (#1339)
   @Inject
   private ITraceSet traceSet;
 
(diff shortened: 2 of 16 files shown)


com.avaloq.tools.ddk.xtext.format.generator/src/com/avaloq/tools/ddk/xtext/format/generator/FormatFragment2.java
```diff
@@ -164,6 +164,7 @@ protected XtendFileAccess doGetXtendStubFile() {
       protected void appendTo(final TargetStringConcatenation builder) {
         builder.append("import com.avaloq.tools.ddk.xtext.formatting.ExtendedLineEntry");
         builder.newLine();
+        // CPD-OFF — parallel Xtend/Java formatter-stub emission, kept explicit (#1339)
         builder.append("import java.util.List");
```
joaodinissf, line 168: I will coalesce these in a follow-up PR.

rubenporras (approved):

PR #1435, opened 2026-06-18
feat(builder): workspace-wide master switch to disable DDK code generation on workspace builds

joaodinissf:
## What

One master checkbox — **"Disable DDK code generation on workspace builds (all DDK languages)"** — shown identically at the top of each meta-DSL's **Compiler** preference page (Check, Scope, Export, Format). It is a single shared workspace preference: toggling it on any page mirrors instantly to the other pages, and while it is active the per-language options below it are grayed out.

## Why

The meta-DSL builder participants regenerate committed artifacts during workspace builds. For DDK contributors who are not currently editing `.check`/`.scope`/`.export`/`.format` sources — especially on slower machines, where clean/full builds re-run generation across every catalog and model in the workspace — that is avoidable build time. The switch turns regeneration off for all four languages in one click; on-demand regeneration via the `Generate*.mwe2` workflows is unaffected.

## Design

- **Default ENABLED** — the preference is a *disable* flag, so the native boolean default (`false` = enabled) holds in every context (IDE, headless, tests) without a preference initializer. Untouched workspaces behave exactly as before.
- **Core:** `BuilderParticipantSettings` (new, `com.avaloq.tools.ddk.xtext.builder`) reads/writes the shared instance-scope node `com.avaloq.tools.ddk`. `ConditionalBuilderParticipant.build()` gates on it; Check and Format check it at the top of their custom `build()` overrides; Scope and Export now bind trivial `ConditionalBuilderParticipant` subclasses in place of the stock `BuilderParticipant` so they are covered too.
- **UI:** `BuilderParticipantMasterSwitch` (new, `com.avaloq.tools.ddk.xtext.ui`) wraps each configuration block's stock content under the master checkbox. Live page instances mirror the pending (unsaved) state instantly, so the four checkboxes always agree and no page can save a stale value over another page's change; nothing persists before Apply, Cancel discards. The gray-out captures each control's own enablement first and restores it afterwards, so enablement the stock page manages (e.g. dependent options) is never clobbered — with the switch untouched, stock pages render pixel-identical. Project property pages are unchanged (the switch is workspace-wide).
- **Re-enabling** generation via Apply/OK schedules a full rebuild (the stock build job), so generated artifacts catch up on edits and deletions made while the switch was off.
- **Explicit `CLEAN` builds are exempt from the gate**: while the switch disables regeneration, a Project → Clean still deletes the generated artifacts (which then stay deleted until re-enabling). Note this is deliberately stricter than the stock per-language "Compiler is activated" pref, which skips the participant wholesale — clean-output included.
- **Relation to the stock per-language "Compiler is activated"** checkbox: that preference already gates a *single* language's participant and remains fully functional; the master switch visibly supersedes it (grays it out) and covers all DDK languages at once.
- **Scope note:** the gate covers every `ConditionalBuilderParticipant` descendant that delegates to `super.build(…)`, including participants of downstream languages built on this class; `build()` overrides that do not delegate must check `isBuilderParticipantEnabled()` themselves (documented in the Javadoc).

## Verification

- Full local gate on this head: `mvn -T 3C clean verify checkstyle:check pmd:check pmd:cpd-check spotbugs:check` → BUILD SUCCESS, 358 tests / 0 failures, all static analysis clean.
- Manually validated on a patched Eclipse 2026-06 install (Xtext 2.43.0, the current target platform): the checkbox renders on all four Compiler pages, cross-page mirroring works before Apply, gray-out and persistence round-trip correctly.
- Adversarially audited: the default path (switch untouched) is behavior-preserving at the builder runtime — including the Scope/Export rebinding — and the two audit findings (stock dependent-control enablement being force-enabled; no rebuild after re-enabling) are fixed in this revision.


joaodinissf:
Closing. Manual verification on a patched Eclipse 2026-06 / Xtext 2.43 install showed the feature works as designed (the switch renders on all four Compiler pages, mirrors across them, and gates generation), but the investigation surfaced two decisive facts:

1. The stock per-language **"Compiler is activated"** checkbox on the same Compiler pages already disables each DDK builder participant — `BuilderParticipant.build()` returns early via `isEnabled()` before any generation. A DDK-wide switch therefore only aggregates four existing checkboxes.
2. The participants do near-zero work when no meta-DSL files changed, so the build-time benefit for developers not editing `.check`/`.scope`/`.export`/`.format` sources is marginal — while the switch's edge semantics (Project > Clean while disabled, stale-artifact catch-up on re-enable) add real maintenance surface.

Too much code for too little benefit given the current review backlog. To skip meta-DSL regeneration today: uncheck *Compiler is activated* on the language's Compiler preference page (per language, workspace- or project-scoped).

*Drafted by Claude Code; reviewed and approved by the author.*

joaodinissf:
Closing again, for good. Reopened to reconsider the pain point, but the July conclusion stands: each DDK language's stock "Compiler is activated" checkbox already disables its builder (verified on master for Check, Scope, Export and Format), incremental builds do almost no work when no DSL files changed, and a DDK-wide switch would also reach downstream languages built on ConditionalBuilderParticipant while leaving other compilers untouched. Making it correct (Apply/OK lifecycle, cleanup of deletions made while disabled, UI tests) is too much surface for a convenience. To skip regeneration, untick "Compiler is activated" per language; it can also be set per project and shared via .settings.

🤖 Drafted with Claude Code

PR #1442, opened 2026-06-21
docs: add Javadoc to public and protected API

joaodinissf:
## What

Documents the public and protected **methods** that a `MissingJavadocMethod` audit surfaced — **51 methods across 21 files**, grouped into four area commits (check, checkcfg, xtext, samples). The diff is **pure additions** (+327, 0 deletions): only Javadoc comments; no code or signatures touched.

The audit rule itself (`MissingJavadocMethod`, public+protected, methods only, trivial-accessor and `@Override`/`@Inject` exemptions) is **not enabled here** — checkstyle cannot see `.xtend` files, so the ongoing Xtend-to-Java migration exposes previously-invisible undocumented API with every merged wave, and the gate would go red after each one. Enablement is deferred to #1466, after the migration completes.

## How the docs were written

Each flagged method gets a **simple, accurate** Javadoc: a one-sentence summary plus the `@param`/`@return`/`@throws` tags the existing content rules require. Descriptions are terse and derived from the signature/obvious behaviour — no speculation, no boilerplate name-restatement. Example:
```java
/**
 * Returns whether references held by the given object via the given feature should be indexed.
 *
 * @param from
 *          the object holding the reference
 * @param eReference
 *          the reference feature to test
 * @return {@code true} unless the reference is a containment or container reference
 */
protected boolean isIndexable(final EObject from, final EReference eReference) { ... }
```

## Verification

- Full local gate (verify + checkstyle + pmd + cpd + spotbugs) green on current master.
- The added Javadoc is validated by the existing `JavadocMethod` content rules.

### DIFF
--- a/com.avaloq.tools.ddk.check.core/src/com/avaloq/tools/ddk/check/CheckRuntimeModule.java
+++ b/com.avaloq.tools.ddk.check.core/src/com/avaloq/tools/ddk/check/CheckRuntimeModule.java
@@ -53,6 +53,11 @@ public class CheckRuntimeModule extends com.avaloq.tools.ddk.check.AbstractCheck
     return CheckBatchLinkableResource.class;
   }
 
+  /**
+   * Binds the resource storage facade.
+   *
+   * @return the resource storage facade class
+   */
   public Class<? extends IResourceStorageFacade> bindIResourceStorageFacade() {
     return CheckBatchLinkableResourceStorageFacade.class;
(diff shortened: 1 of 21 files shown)


rubenporras (approved):

rubenporras:
something is wrong in the CI/CD setup, we should enforce this with PMD or Checkstyle, in case you want to chase it further

PR #1446, opened 2026-06-24
fix: migration faithfulness follow-ups (generator byte-drift + null-tolerance)

joaodinissf:
## Context

Follow-up to the Xtend→Java migration campaign. After the migrations merged, every migrated file was audited for **generated-output byte-drift** (where a code generator's emitted text must be byte-identical to what Xtend produced) and general faithfulness. The audit compared each migrated `.java` against the authoritative `xtend-gen` (the Xtend compiler's own output), with executable `StringConcatenation` harnesses for the non-trivial generators. This PR carries the two **generator-code** fixes that surfaced; both are verified byte-identical / behaviour-faithful to `xtend-gen`.

## Fixes

**1. `AnnotationAwareContentAssistFragment2` — restore continuation-line indentation**
The generated `Abstract…ProposalProvider` was built with single-arg `append(value)`; when an assignment's terminal is a multi-line `Alternatives`, the continuation lines lost their indentation versus the Xtend output (Xtend uses two-arg `append(value, indent)`, which re-indents every continuation line). Restored the Xtext relative + two-arg-reindent pattern.
*Verification:* an executable `StringConcatenation` harness ran both the `xtend-gen` builder and the migrated builder over 10 input shapes (empty / single- & multi-line leaf / options / keywords / stub & non-stub / combined): **7/10 → 10/10 IDENTICAL** after the fix.

**2. `FormatJvmModelInferrer` — null-tolerant `Arrays.asList` in dispatcher throws**
The synthesized `@Dispatch` fall-through throws built the "Unhandled parameter types" message with `List.of(...)`, which throws `NullPointerException` on a null element; the Xtend-generated code used `Arrays.<Object>asList(...).toString()` (null-tolerant). On a defensive null path this changed the thrown exception type. Restored `Arrays.<Object>asList`, matching `xtend-gen`.

## Two additional minor observations (intentionally not changed here)

- **`CheckCfgScopeProviderTest`**: the migration replaced an explicit `new NullPointerException("Got null context model")` with `new IllegalStateException(...)` (same message, same test outcome). Kept as-is — an explicit `IllegalStateException` is preferable to throwing `NullPointerException`; reverting purely for byte-faithfulness would be a small quality regression.
- **`CheckModelUtil.modelWithContexts`** (test helper): emits a trailing `"\n    "` the Xtend `«FOR»` did not. Its only caller is an `@Disabled` test, so it is never exercised; deferred to avoid churn.

### DIFF
--- a/com.avaloq.tools.ddk.xtext.format/src/com/avaloq/tools/ddk/xtext/format/jvmmodel/FormatJvmModelInferrer.java
+++ b/com.avaloq.tools.ddk.xtext.format/src/com/avaloq/tools/ddk/xtext/format/jvmmodel/FormatJvmModelInferrer.java
@@ -11,6 +11,7 @@
 package com.avaloq.tools.ddk.xtext.format.jvmmodel;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.LinkedList;
@@ -1431,7 +1432,7 @@ public class FormatJvmModelInferrer extends AbstractModelInferrer {
     } else if (format != null) {
       _infer(format, acceptor, isPreIndexingPhase);
     } else {
-      throw new IllegalArgumentException("Unhandled parameter types: " + List.of(format, acceptor, isPreIndexingPhase));
+      throw new IllegalArgumentException("Unhandled parameter types: " + Arrays.<Object>asList(format, acceptor, isPreIndexingPhase));
     }
   }
 
@@ -1441,7 +1442,7 @@ public class FormatJvmModelInferrer extends AbstractModelInferrer {
     } else if (rule instanceof WildcardRule wildcardRule) {
       return _getLocatorActivatorSuperType(formatConfiguration, wildcardRule);
     } else {
-      throw new IllegalArgumentException("Unhandled parameter types: " + List.of(formatConfiguration, rule));
+      throw new IllegalArgumentException("Unhandled parameter types: " + Arrays.<Object>asList(formatConfiguration, rule));
     }
   }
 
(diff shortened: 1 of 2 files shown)


com.avaloq.tools.ddk.xtext.format/src/com/avaloq/tools/ddk/xtext/format/jvmmodel/FormatJvmModelInferrer.java
```diff
@@ -1473,7 +1474,7 @@ public String getRuleName(final EObject rule) {
     } else if (rule != null) {
       return _getRuleName(rule);
     } else {
-      throw new IllegalArgumentException("Unhandled parameter types: " + List.of());
+      throw new IllegalArgumentException("Unhandled parameter types: " + Arrays.<Object>asList());
```
rubenporras:
```suggestion
      throw new IllegalArgumentException("Unhandled parameter types");
```

com.avaloq.tools.ddk.xtext.format/src/com/avaloq/tools/ddk/xtext/format/jvmmodel/FormatJvmModelInferrer.java
```diff
@@ -1489,7 +1490,7 @@ public String getDirectiveName(final EObject directive) {
     } else if (directive != null) {
       return _getDirectiveName(directive);
     } else {
-      throw new IllegalArgumentException("Unhandled parameter types: " + List.of());
+      throw new IllegalArgumentException("Unhandled parameter types: " + Arrays.<Object>asList());
```
rubenporras:
```suggestion
      throw new IllegalArgumentException("Unhandled parameter types");
```

com.avaloq.tools.ddk.xtext.format/src/com/avaloq/tools/ddk/xtext/format/jvmmodel/FormatJvmModelInferrer.java
```diff
@@ -1549,7 +1550,7 @@ public CharSequence elementAccess(final Object grammarElementLookup) {
     } else if (grammarElementLookup != null) {
       return _elementAccess(grammarElementLookup);
     } else {
-      throw new IllegalArgumentException("Unhandled parameter types: " + List.of());
+      throw new IllegalArgumentException("Unhandled parameter types: " + Arrays.<Object>asList());
```
rubenporras:
```suggestion
      throw new IllegalArgumentException("Unhandled parameter types");
```

rubenporras (changes requested):

rubenporras (approved):

PR #1473, opened 2026-07-19
docs: harden Xtend-to-Java migration fidelity rules

joaodinissf:
## What

Hardens `.agents/skills/xtend-to-java` with migration-fidelity rules verified against Xtext 2.43.0, fresh `xtend-gen`, and current repository migrations.

## Rules and fixes

- Preserve both `JvmTypesBuilder.operator_add` null-skipping overloads at inferrer `EList +=` sites; nullable builders and helpers require guards or `Objects::nonNull`.
- Preserve `IterableExtensions.toSet` encounter order with `LinkedHashSet`, and check whether its return-existing-`Set` aliasing is observable.
- Preserve Xtend checked-exception signatures and throwable identity without copying compiler scaffolding. Keep explicit catches specific; catch only the narrow checked types Java requires. When a checked exception cannot be declared (overrides, compatibility-sensitive APIs), do not pick a workaround mechanically — obtain explicit review for the boundary strategy. Do not catch `Throwable`, `Exception`, or `RuntimeException` merely to imitate `xtend-gen`; a broad catch is allowed only when the invoked API declares that type and no narrower form compiles.
- Choose charsets from the data contract. Use Eclipse file metadata or UTF-8 for repository-owned UTF-8 text; do not guess UTF-8 for opaque external bytes.
- Align the quick reference, validation checklist, known pitfalls, and worked example; replace stale line-number references with stable method references.

## Why

The `operator_add` rule prevents the regression fixed in #1474: `FormatJvmModelInferrer.inferConstants` used a bare `EList.add` for a nullable producer and threw where Xtend skipped null. The ordered-set rule protects generator output in the remaining Wave 4 scope/export migrations.

## Validation

- `git diff --check`
- all relative Markdown links resolve
- Xtext 2.43.0 source, bytecode, and compiler assertions reviewed
- `mvn -f ./ddk-parent/pom.xml -pl :com.avaloq.tools.ddk.xtext.scope -am -DskipTests -T 1 compile --batch-mode` — `BUILD SUCCESS`

Skill-only Markdown change; no production or build-configuration changes.


.agents/skills/xtend-to-java/workflow/validation-checklist.md
```diff
@@ -123,14 +131,14 @@ Every rule below is a hard gate.
...
-- [ ] All `+=` / `-=` on collections converted to `.add()` / `.addAll()` / `.remove()`.
+- [ ] All `+=` / `-=` on ordinary collections converted to `.add()` / `.addAll()` / `.remove()`; inferrer `EList +=` sites preserve `JvmTypesBuilder.operator_add` null-skipping.
 - [ ] All `::` static-access converted.
 - [ ] Semicolons added to every statement.
 - [ ] Explicit visibility on every class, method, and field.
 - [ ] All imports updated (no wildcards, no unused, correct order).
 - [ ] All comments and class/member Javadoc preserved exactly.
 - [ ] Copyright header is the exact Avaloq banner — **replacing** any generated-stub or Javadoc-style header the source had (not preserved from source).
-- [ ] Checked exceptions handled (`throws` clause or `try`/`catch` with specific types).
+- [ ] Checked exceptions preserve the original signature and throwable identity using only narrow catches; `Exceptions.sneakyThrow` is used where required.
```
rubenporras: I would not use `Exceptions.sneakyThrow`

rubenporras (approved):

PR #1474, opened 2026-07-19
fix(format): guard against null members in FormatJvmModelInferrer.inferConstants

joaodinissf:
## The bug

`FormatJvmModelInferrer.inferConstants` adds `createConstant(format, c)` to the inferred JVM model **unconditionally**:

```java
for (final Constant c : FormatGeneratorUtil.getAllConstants(format)) {
  it.getMembers().add(createConstant(format, c));   // createConstant can return null
}
```

`createConstant` returns `null` for a **value-less constant** (its `if (stringValue) … else if (intValue) … return null` fall-through — reachable for a constant in an incomplete/error state during editing). Adding that null to the `JvmMember` `EList` throws `IllegalArgumentException: The 'no null' constraint is violated`.

## Root cause — a migration regression

The original Xtend was `members += allConstants.map[createConstant]`, which compiles to `JvmTypesBuilder.operator_add`, and **`operator_add` silently skips null elements**. The Xtend→Java migration translated `+=` to a bare loop-`add`, which does not skip nulls — so the null-skip semantics were lost. The sibling `inferRules` in the same file kept the null-skip, making this an inconsistent one-off.

## The fix

Guard the add (equivalent to `operator_add`'s null-skip / an `IterableExtensions.filterNull`):

```java
final JvmMember member = createConstant(format, c);
if (member != null) {
  it.getMembers().add(member);
}
```

## Verification

- New `FormatJvmModelInferrerTest` pins the behaviour: a value-less `Constant` must not yield a null member or throw.
- **A/B proven** through the full test harness: the test **fails on the pre-fix code** (`IllegalArgumentException: The 'no null' constraint is violated`) and **passes with the guard**. Full local gate: BUILD SUCCESS, 359 tests / 0 failures, all static analysis clean.

Found by an archaeology audit of the merged Xtend→Java migrations. The skill rule that prevents this class of defect is added in a separate PR (#1473, `rules/10 §10.5` — `operator_add` skips nulls).

### DIFF
--- a/com.avaloq.tools.ddk.xtext.format.test/src/com/avaloq/tools/ddk/xtext/format/jvmmodel/FormatJvmModelInferrerTest.java
+++ b/com.avaloq.tools.ddk.xtext.format.test/src/com/avaloq/tools/ddk/xtext/format/jvmmodel/FormatJvmModelInferrerTest.java
@@ -1,0 +1,74 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Avaloq Group AG and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Avaloq Group AG - initial API and implementation
+ *******************************************************************************/
+package com.avaloq.tools.ddk.xtext.format.jvmmodel;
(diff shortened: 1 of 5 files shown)


rubenporras:
Can you rebase this commit?

rubenporras:
The maven verify passed, so our baseline check is not working, I will need to investigate, or if you can, please, be my guest :)

rubenporras (approved):

PR #1511, opened 2026-09-01
test: apply the SWTBot widget timeout suite-wide

joaodinissf:
## Summary

- apply a 20 s SWTBot widget timeout from a static initializer in `SwtWorkbenchBot`, the entry point every UI test already constructs
- keep an explicit `-Dorg.eclipse.swtbot.search.timeout` authoritative: the initializer only raises the value when the property is absent
- delete `SwtBotUtil` and the `initializePreferences` / focus-policy chain in `CoreSwtbotTools`, none of which had a caller
- bump `com.avaloq.tools.ddk.test.ui` to 17.4.0 because public API is removed

## Why

Every SWTBot wait in the suite has been running at the library default of 5 s. DDK wrote down a 20 s value in `CoreSwtbotTools.initializePreferences()` in 2017, but it was only ever invoked by the JUnit 4 recording runners and `AbstractUiTest`. Those served three abstract bases with no concrete subclass in this repository, and they were removed in March and May 2026. The 10 s variant in `SwtBotUtil` had no caller at all.

The 5 s default sits below the Problems view's deferred refresh: Eclipse's `MarkerUpdateTimer` re-arms the view update 10 s out whenever marker deltas keep arriving after a build, and up to 30 s while a build is still running. A 5 s lookup can therefore observe zero rows from a view that is merely late. That was the amplifier behind the `testBulkApplyingQuickfix` flake addressed in #1510, which had to thread its own 30 s bound through nine call sites. This change removes the amplifier for every UI test; the 30 s per-site bound in `ProblemsViewTestUtil` stays because 20 s does not cover the while-building case.

Not applied: playback delay, keyboard layout and type interval from the old initializer. Those change input behaviour for every test and are a separate decision.

Cost: negative lookups take longer to fail. The existing absence checks go through the 64 ms `ImpatientSwtWorkbenchBot`, so they are unaffected. The two retry attempts in `DynamicContextActionUiTestUtil.clickContextMenu` wait the full timeout before falling through, so a genuinely missing context menu now costs 60 s instead of 15 s before the test fails.

## Local validation

- full CI-equivalent clean reactor on `b7bc173ba` with `verify checkstyle:check pmd:pmd pmd:cpd pmd:check pmd:cpd-check spotbugs:check`: passed, 5:30 min wall clock
- aggregate suite: 354 tests, 0 failures, 0 errors, 2 skipped
- `CheckQuickfixTest`: 3 tests passed, `testBulkApplyingQuickfix` 6.0 s


com.avaloq.tools.ddk.test.ui/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: com.avaloq.tools.ddk.test.ui
 Bundle-SymbolicName: com.avaloq.tools.ddk.test.ui;singleton:=true
-Bundle-Version: 17.3.3.qualifier
+Bundle-Version: 17.4.0.qualifier
```
rubenporras, line 5: this breaks compatibility, so it should be 18.0.0, right?

rubenporras (changes requested):
Besides the comment on the versioning, have you checked downstream projects? I see the API we removed is used by them

joaodinissf:
You're right on both counts, and following up the second one invalidated the
premise of this PR, so I'm closing it.

On downstream usage: confirmed. `SwtBotUtil` and the `CoreSwtbotTools`
focus-policy members are dead only within DDK. Their callers here were the
JUnit 4 runners removed in the JUnit 4 infrastructure cleanup, and downstream
kept calling them directly from its own setup. Removing them was wrong
regardless of the version number, so the 17.4.0 vs 18.0.0 question is moot.

The timeout change itself doesn't hold up either. I justified raising
`SWTBotPreferences.TIMEOUT` as outlasting the Problems view's deferred marker
refresh, but every Problems view wait in `ProblemsViewTestUtil` and
`CheckQuickfixTest` already passes an explicit `ASYNC_UPDATE_TIMEOUT` to
`waitUntil`, and that overload ignores the global preference. The global value
only governs implicit widget lookups, and there's no failure attributable to
those: master has run at the SWTBot default since the cleanup and is green. The
20s figure came from a 2017 constant with no measurement behind it, and raising
it would quadruple the time every genuine lookup failure takes to report.

The pattern already in the codebase is the right one, and #1510 follows it:
bound each async wait explicitly at the call site.

---
🤖 Written by Claude, reviewed and sent by @joaodinissf

PR #1512, opened 2026-09-02
fix: remove wrapping of Post/PreConditionViolation

gabhernando31:
Removed wrapping of Post/PreConditionViolation in
AbstractTestStep to allow AssertionFailedException to be thrown int SystemTests.

Issue: ACF-14641


com.avaloq.tools.ddk.test.core/src/com/avaloq/tools/ddk/test/core/AbstractTestStep.java
```diff
@@ -151,7 +151,7 @@ protected void runBefore() {
         // CHECKSTYLE:CHECK-OFF IllegalCatch // exception is wrapped
       } catch (Throwable t) {
```
rubenporras: we can remove the try / catch, it does not serve any purpose now

com.avaloq.tools.ddk.test.core/src/com/avaloq/tools/ddk/test/core/AbstractTestStep.java
```diff
@@ -170,7 +170,7 @@ protected void runAfter() {
       // CHECKSTYLE:CHECK-OFF IllegalCatch // exception is wrapped
     } catch (final Throwable t) {
       // CHECKSTYLE:CHECK-ON IllegalCatch
-      throw new PostconditionViolation(t);
```
rubenporras, line 173: we can remove the try / catch, it does not serve any purpose now

rubenporras (changes requested):
We also discussed removing PostconditionViolation, Could we do that as well?

rubenporras (changes requested):
Also remove `PostconditionViolation` and `ContractViolation`

gabhernando31:
> Also remove `PostconditionViolation` and `ContractViolation`

Should We remove "PreConditionViolation" too? since it also extends "ContractViolation".
But PreConditionViolation is being used in CompoundStep.java

rubenporras:
> > Also remove `PostconditionViolation` and `ContractViolation`
> 
> Should We remove "PreConditionViolation" too? since it also extends "ContractViolation". But PreConditionViolation is being used in CompoundStep.java

Yes, I think we can replace it with org.opentest4j.TestAbortedException

rubenporras (dismissed):

rubenporras (approved):

PR #1515, opened 2026-09-04
fix: satisfy PMD 7.27.0 across the reactor

joaodinissf:
## Summary

Upgrade PMD from 7.26.0 to 7.27.0 and address only the reproduced warnings, without optional refactoring. The version bump from Dependabot #1514 is included; close #1514 after this PR merges.

## Scope

The unfixed PR base with PMD 7.27 produces 24 violations across 19 Java files: 4 InstantiableUtilityClass, 13 ProtectedMemberInFinalClass, 4 ReturnEmptyCollectionRatherThanNull, and 3 UnusedReturnValue. Report generation across all modules was necessary because the normal fail-at-end check skips downstream modules after the initial failures.

- Preserve four exported utility classes' non-final declarations and protected constructors with targeted compatibility suppressions.
- Narrow non-overriding implementation members to private or package-private; preserve injection and same-package constructor access.
- Preserve the three existing exception-based existence searches with documented UnusedReturnValue suppressions. No lambda conversions, getter caching, or search-strategy changes are included.
- Remove redundant always-null local branches in the two validation-element collection methods. Preserve the meaningful proxy-node null sentinel with a targeted suppression.
- Initialize tableItems' result to an empty list, consistent with its documented non-null return contract. This warning-related change can replace a null result with an empty list if the table becomes empty after the wait. The pre-existing per-row overwrite bug remains separate in #1516.
- Retain the private loading helper; only its visibility changes.
- Include necessary plugin/feature version metadata. Remove Index's constructor `@return` tag because local Checkstyle rejects it; unrelated comment cleanup is excluded.

Optional simplifications and their behavioral tests are being submitted in a separate dependent draft PR. They are not prerequisites for merging this PMD update.

## Validation

The warning audit used Java 21 and CI's exact Maven PMD/CPD goals on macOS. The original five blocking CI findings reproduced, and report-only execution exposed all 24 findings across 62 PMD reports; 62 CPD reports contained no duplicates.

Full local verification and quality checks passed on head `296d03baa`: Maven verify, Checkstyle, PMD, CPD, and SpotBugs. The aggregate suite reported 354 tests, 0 failures, 0 errors, and 2 skipped. This was Java 21 on macOS, using the repository's macOS UI profile. Fresh remote checks are still required on this head; previous-head CI is not evidence for this revision.

Keep draft with no reviewer requests pending validation and review.

PR #1521, opened 2026-09-05
refactor: simplify existence checks and object-loading helper

joaodinissf:
## Summary

Optional simplifications deliberately separated from PMD-only #1515. This PR is based on `codex/pmd-7.27-base`, an upstream mirror of the narrowed #1515 head, so the review diff contains only this follow-up. After #1515 merges, retarget this PR to master and rerun checks before merging.

## Changes

- Replace exception-driven existence searches with boolean predicates; simplify both catalog checks with lambdas and one local catalog reference while retaining completeness checks before identity comparison.
- Replace marker-help key lookup plus predicate scanning with direct multimap entry membership. Xtext pairs use value-based equality and hashing.
- Inline the private loading helper, removing its unused parameter, unnecessary exception declaration, and suppression while preserving adapter attachment behavior.
- Add six tests in the existing aggregate suite for matching entries/default check kind, missing context with equal issue count, wrong marker type, wrong issue code, null marker type, and null issue code.
- Advance affected runtime plugin/feature versions independently of #1515.

## Behavioral boundaries

This is not claimed to be strictly exception-equivalent. The former searches caught NoSuchElementException from the entire find call, including predicate evaluation; the new checks do not intentionally swallow predicate exceptions. Null-valued marker-help pair components are non-matches against expected non-null values instead of potentially throwing NullPointerException. The new regression tests cover these null-valued entries at the Eclipse model level. Normal Check kinds default to FAST, including when set to null, and successfully generated issue-code values are non-null strings.

## Validation

Local full verification and quality checks passed on head `21e9bae62`: Maven verify, Checkstyle, PMD, CPD, and SpotBugs using Java 21 on macOS. The aggregate suite reported 360 tests, 0 failures, 0 errors, and 2 skipped. All 11 CheckMarkerHelpExtensionTest cases passed, including the six new cases. They run through the existing aggregate suite; no separate test runtime configuration is added. This draft must receive its own fresh green CI before merge.

Keep draft with no reviewer requests. Review independently from #1515; do not merge before its dependency.

### DIFF
--- a/com.avaloq.tools.ddk.check.ui.test/src/com/avaloq/tools/ddk/check/ui/test/builder/CheckMarkerHelpExtensionTest.java
+++ b/com.avaloq.tools.ddk.check.ui.test/src/com/avaloq/tools/ddk/check/ui/test/builder/CheckMarkerHelpExtensionTest.java
@@ -11,6 +11,7 @@
 package com.avaloq.tools.ddk.check.ui.test.builder;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.util.List;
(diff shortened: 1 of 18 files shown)


joaodinissf:
Superseded by #1531 — same commit (`21e9bae6`), re-opened from an `upstream` branch and now natively stacked on #1530.

🤖 Written by Claude, reviewed and sent by @joaodinissf

PR #1522, opened 2026-09-06
fix(scope): defer provider inference during model linking

joaodinissf:
Scope JVM inference can fail while a grammar is being linked, leaving the Scope resource initialized without its provider types. This fixes the lifecycle regression introduced by #1405 (merged 19 August 2026), present in DDK 19.1.

The migration rendered Java bodies eagerly before accepting either inferred type. A resource-content read during grammar linking could therefore reach `GenModelUtilX.literalIdentifier(EClass)` while the classifier had no EPackage. Delaying only bodies is insufficient: helper names and inheritance merging also resolve model types, and an early member query can leave incomplete declarations cached.

The fix accepts both type shells first, uses demand-driven member initialization, checks includes/signatures before inheritance traversal, and renders bodies during emission with fresh model-specific generators and restored resource-loader context. A resource whose earlier member initialization was incomplete is inferred again before generation; if it remains incomplete, no Java is written. Existing resolved helper names and generation algorithms are preserved.

The root-cause analysis and portable public Eclipse reproducer are included under [`reproducers/scope-inference`](https://github.com/dsldevkit/dsl-devkit/tree/codex/fix-scope-inference/reproducers/scope-inference). The four tests cover resource-root traversal, early member demand and repair, refusing generation while unresolved, and recovery of inherited reference helpers/injections. They assert actual method declarations, not merely dispatch-call text. Two independent adversarial reviews informed the recovery design; the final review found no remaining blocking defect in this scope.

Validation:

- GitHub CI passed for commit `cdb5869a6850b7867696806ce1fe05017dd5f8d7`: `maven-verify`, `pmd`, `checkstyle`, and `line-endings`. The local UI timeout noted above did not prevent a successful CI verification run.

- Public DDK 19.1 target: all four regression tests fail as expected.
- Fixed p2 target: the same four tests pass, including from a freshly extracted reproducer ZIP.
- Four generated provider Java classes compile with Java 21.
- Full suite, twice (including a fresh test workspace): 358 tests, zero assertion failures, two skipped, one error in the existing `testBulkApplyingQuickfix` UI test waiting for the “Select a fix” table. All four new Scope tests passed in both runs. This UI timeout remains unresolved.
- All repository Checkstyle, PMD, CPD, and SpotBugs checks passed.

The fixture stages unavailable model objects and proves this DDK failure and recovery mechanism. It does not reproduce the entire private downstream grammar/Export loading sequence or establish why its model became unavailable. Local runs were on macOS/JDK 21; Windows verification remains to be done. This PR contains only the Scope fix.

### DIFF
--- a/com.avaloq.tools.ddk.xtext.generator.test/src/com/avaloq/tools/ddk/xtext/generator/expression/ScopeInferenceTest.java
+++ b/com.avaloq.tools.ddk.xtext.generator.test/src/com/avaloq/tools/ddk/xtext/generator/expression/ScopeInferenceTest.java
@@ -1,0 +1,130 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Avaloq Group AG and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Avaloq Group AG - initial API and implementation
+ *******************************************************************************/
+package com.avaloq.tools.ddk.xtext.generator.expression;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+
+import org.eclipse.emf.common.util.URI;
+import org.eclipse.emf.ecore.EClass;
+import org.eclipse.emf.ecore.EcoreFactory;
+import org.eclipse.emf.ecore.EcorePackage;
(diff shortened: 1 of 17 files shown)


joaodinissf:
A synthetic reproducer for the Scope inference regression addressed by this PR is ready; I’ll attach the ZIP shortly.

With JDK 21, Maven, and Python 3 installed, extract the ZIP and run:

```text
python prepare.py
mvn clean verify
```

Against public DDK 19.1, all four tests fail as expected. The same tests pass against the fixed build; the README includes instructions for selecting that target and running in Eclipse.

The tests cover incomplete model references and includes, early member access, and recovery of generated provider methods. The ZIP was validated from a fresh extraction on macOS. Windows verification remains outstanding.

This reproduces the inference failure mechanism, rather than the complete downstream build slowdown.

joaodinissf:
[scope-inference-reproducer.zip](https://github.com/user-attachments/files/31909829/scope-inference-reproducer.zip)

joaodinissf:
Closing as likely not needed after https://github.com/dsldevkit/dsl-devkit/pull/1526
Available for reference if the problem recurs.

PR #1523, opened 2026-09-07
feat: drop the valid language

rubenporras:
since all the downstream projects have already moved away and there is no known consumer


joaodinissf (approved):
The bundle removal and build/feature integration look correct. I will address the remaining Valid regeneration workflow references and stale README/CODEOWNERS entries in a follow-up PR. Shared validation support used by Check DSL will remain intact.

PR #1524, opened 2026-09-07
fix: remove obsolete Valid generation workflow

joaodinissf:
PR #1523 removes the Valid language bundles, but the aggregate generation workflow still invokes Valid and its dedicated workflow references the deleted grammar and metamodels. Remove that invocation and `GenerateValid.mwe2`, together with the stale README module row and CODEOWNERS entry.

Follow-up to merged PR #1523, based on updated `master`. The diff contains only the four cleanup files.

Shared validation support remains unchanged, including `ValidCompositeEValidator`, which Check DSL still extends.

Validation: repository reference scan confirms no remaining removed-bundle references or Valid workflow invocation; shared Check validator inheritance verified; `git diff --check` and the repository line-ending check passed. Maven and Eclipse/MWE2 regeneration were not rerun; the workflow project is outside the Maven reactor.


CODEOWNERS
```diff
@@ -10,9 +10,6 @@
 /com.avaloq.tools.ddk.check.*/**                       @dsldevkit/sca
```
joaodinissf, line 10: @rubenporras I think this file is basically useless now--I'd suggest we either fix this (I don't have permissions) or delete the file, as you prefer.

joaodinissf: Image

rubenporras (approved):

CODEOWNERS
rubenporras: Yes, I think we should remove it

PR #1540, opened 2026-09-11
docs(skill): record jvmmodel-batch learnings in xtend-to-java

joaodinissf:
Records in the `xtend-to-java` skill what the final jvmmodel batch (#1536) taught, and makes the skill repository-agnostic so it can be reused in other Eclipse/Tycho/Xtext projects: commands use `<parent-pom>` / `<target-module>` placeholders, and repository conventions (license header, shared JDT settings) come from the repository itself. Documentation only; no code.

- Tier 4 templates stay on the reference `StringConcatenation` chain when they carry `newLineIfNotEmpty()` after a dynamic value, a two-argument indented append of a possibly multi-line value, `appendImmediate`, or a `StringConcatenationClient` value. `append(null)` appends nothing, so a nullable `String` moved into `.formatted()` wraps in `Strings.emptyIfNull`, other nullable values in `Objects.toString(value, "")`.
- `JvmTypeReferenceBuilder.typeRef` is provably non-null; `to*` builders with a nullable name stay guarded.
- Dispatcher case order is taken from `xtend-gen`, never from source order; the terminal else-throw stays.
- Checkstyle `JavadocMethod` vs "never invent Javadoc": complete the missing tag from a sibling overload.
- `xbase.lib` types stay in public signatures and callee-demanded callbacks.
- `BooleanExpressionComplexity` vs `SimplifyBooleanReturns`: guard clauses or a named local, no suppression.
- PMD `StringToString`: `Integer.toString(x)` is neutral; dropping `toString()` on a `String` needs a non-null proof, otherwise `Objects.requireNonNull`.
- Module-scoped gate commands include the target-platform module; clear the stale `xtend-gen` output before the first compile; `diff -r -x '.*'` for the freshness check.
- If the build runs `compare-version-with-baselines` (`tycho-p2-extras-plugin`): a separate `build:` bump commit for the bundle, then features and update site only where they also equal the baseline.
- `.project` Xtext builder/nature stay only while the module still has Xtext-language resources; otherwise they were Xtend's and go.


.agents/skills/xtend-to-java/workflow/formatting-and-commit.md
```diff
@@ -95,6 +95,19 @@ Known trade-offs (accepted):
 Do not split the translate step per file — every intermediate commit before the last file
 would be broken anyway, so per-file translate commits only multiply the broken range.
 
+## Tycho baseline bump
```
rubenporras, line 88: I would not add this, as it is different for downstream projects

.agents/skills/xtend-to-java/workflow/multi-file-batch.md
```diff
@@ -34,10 +34,14 @@ Batch files within the **same module** together where possible — they share Ma
 
 After every batch:
 
-1. **Compile gate**: `mvn -pl :<module> -am -DskipTests compile -f ./ddk-parent/pom.xml` — must pass.
+1. **Compile gate**: `mvn -pl :ddk-target,:<module> -am -DskipTests compile -f ./ddk-parent/pom.xml` — must pass.
```
rubenporras: I would not add this, as it is different for downstream projects

.agents/skills/xtend-to-java/workflow/multi-file-batch.md
```diff
@@ -34,10 +34,14 @@ Batch files within the **same module** together where possible — they share Ma
...
 2. **Test gate**: `mvn verify -f ./ddk-parent/pom.xml --batch-mode --fail-at-end` — must pass.
 3. **Static analysis gate**: `mvn checkstyle:check pmd:check spotbugs:check -f ./ddk-parent/pom.xml` — must pass.
 
+`:ddk-target` must be in every `-pl` list — the target artifact is not in `~/.m2`. Before the first
```
rubenporras: I would not add this, as it is different for downstream projects

.agents/skills/xtend-to-java/workflow/one-file-conversion.md
```diff
@@ -14,11 +14,11 @@ Use this when converting a single `.xtend` to its `.java` counterpart.
 8. **Commit as two steps** — a pure `git mv` rename commit, then an in-place translate commit; see [`formatting-and-commit.md`](./formatting-and-commit.md) §Commit structure. For a single file the `git mv` IS the removal; do not delete+re-add.
 9. **Verify the file compiles:**
    ```bash
-   mvn -pl :<module> -am -DskipTests compile -f ./ddk-parent/pom.xml > mvn-output.txt 2>&1
+   mvn -pl :ddk-target,:<module> -am -DskipTests compile -f ./ddk-parent/pom.xml > mvn-output.txt 2>&1
```
rubenporras: probably this and the others need to be more generic so that it is not so tied with DDK naming, don't you think?

rubenporras (changes requested):

.agents/skills/xtend-to-java/workflow/formatting-and-commit.md
joaodinissf: Agreed. The section now applies only when the build runs the baseline check, and names no DDK modules.

🤖 Drafted with Claude Code

.agents/skills/xtend-to-java/workflow/multi-file-batch.md
joaodinissf: Agreed, replaced with generic placeholders.

🤖 Drafted with Claude Code

.agents/skills/xtend-to-java/workflow/multi-file-batch.md
joaodinissf: Agreed, now phrased as a generic Tycho rule.

🤖 Drafted with Claude Code

.agents/skills/xtend-to-java/workflow/one-file-conversion.md
joaodinissf: Agreed. I went through the whole skill: no DDK-specific names are left apart from the `@Tag` section, which only applies to projects using the DDK test framework.

🤖 Drafted with Claude Code

rubenporras (approved):

